PR #3, opened 2023-10-25
Feature/ele 144 routing

gularsson:
Known 
- [ ] Fix GHA build (inadvertently added Dockerfile and build.yaml, might work :crossed_fingers:)
- [ ] Remove hardcoded svg icons, use `lucide-react` from `elephant-ui`
- [ ] Might have to refactor `NavigationProvider`/`NavigationContext` (hard to test dispatches this way)


src/contexts/NavigationProvider.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+  useLayoutEffect(() => {
+    // Create history state for initial content
+    if (historyState === null) {
+      // TODO: Make router/location hook
+      const currentView = window.location.pathname !== '/'
+        ? window.location.pathname[1].toUpperCase() + window.location.pathname.slice(2)
+        : 'Init'
+      const currentProps = Object.fromEntries(new URLSearchParams(window.location.search))
+      const id = currentProps.id || 'init' // eslint-disable-line @typescript-eslint/strict-boolean-expressions
```
dannelundqvist, line 31: This eslint-disable can be removed **later**, when merged with add-basic-editor-and-yjs branch where I have disabled this stupid rule :)

src/types/index.ts
```diff
@@ -0,0 +1,60 @@
...
+  }
+  component: React.FC<ViewProps>
+}
+
+export interface Registry {
+  get: (key: View) => RegistryItem
+  set: () => void
+}
+
+export interface NavigationState {
```
dannelundqvist, line 33: This is sometimes used without the properties as empty object. If that is correct it should be
```
export interface NavigationState {
  registry?: Registry
  content?: JSX.Element[]
}
```

src/components/NavigationWrapper/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+  const variants = {
+    maximized: 'absolute inset-y-0 left-0 z-10 w-screen h-screen bg-gray-100 basis-full rounded-lg p-2',
+    maxIcon: <svg xmlns="http://www.w3.org/2000/svg" width="18" height="18" viewBox="0 0 24 24" fill="none" stroke="currentColor" strokeWidth="2" strokeLinecap="round" strokeLinejoin="round" className="lucide lucide-minimize-2"><polyline points="4 14 10 14 10 20"/><polyline points="20 10 14 10 14 4"/><line x1="14" x2="21" y1="10" y2="3"/><line x1="3" x2="10" y1="21" y2="14"/></svg>,
+    minimized: 'flex-grow bg-gray-200 basis-full rounded-lg p-2 min-w-max',
+    minIcon: <svg xmlns="http://www.w3.org/2000/svg" width="18" height="18" viewBox="0 0 24 24" fill="none" stroke="currentColor" strokeWidth="2" strokeLinecap="round" strokeLinejoin="round" className="lucide lucide-maximize-2"><polyline points="15 3 21 3 21 9"/><polyline points="9 21 3 21 3 15"/><line x1="21" x2="14" y1="3" y2="10"/><line x1="3" x2="10" y1="21" y2="14"/></svg>
+  }
+
+  return useMemo(() => (
+    <section className={expanded ? variants.maximized : variants.minimized}>
+      <div className="absolute right-0 top-0 p-2"
```
dannelundqvist, line 14: Picky ;) but when making elems multiline, readability is improved when all attribs are on their own line as in below.
```
<div
   className="absolute right-0 top-0 p-2"
   onClick={() => { setExpanded(!expanded) }}
>
      {expanded ? variants.maxIcon : variants.minIcon}
</div>
```

src/components/Link/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+interface LinkProps {
+  children: string
+  to: View
+  props?: Record<string, unknown>
+}
+
+export const Link = ({ children, to, props }: LinkProps): JSX.Element => {
+  const { state, dispatch } = useNavigation()
+  const id = uuid()
+  const linkItem = state.registry.get(to)
```
dannelundqvist, line 14: If changing NavigationState interface to allow empty state (as in other comment), this would be possibly undefined.

src/contexts/NavigationProvider.tsx
```diff
@@ -0,0 +1,109 @@
...
+import { NavigationWrapper } from '@/components/NavigationWrapper'
+import type { ContentState, NavigationState, NavigationAction } from '@/types'
+import { NavigationActionType } from '@/types'
+
+import { useHistory } from '@/hooks'
+
+export const NavigationContext = createContext<{
+  state: NavigationState
+  dispatch: Dispatch<NavigationAction>
+}>({ state: {} as NavigationState, dispatch: () => {} }) // eslint-disable-line @typescript-eslint/consistent-type-assertions
```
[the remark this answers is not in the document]
dannelundqvist: This would have an impact in many other places where this is used though...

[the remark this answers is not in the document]
dannelundqvist: ...or maybe have sane defaults so that it is not empty.

src/components/Link/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+  return (
+    <a
+      className='p-1 hover:font-bold'
+      href={`${linkItem.metadata.path}?id=${id}`}
+      onClick={(event) => {
```
dannelundqvist, line 19: The onClick code should be refactored into it´s own function in this module to enhance readability

src/components/Link/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+        event.preventDefault()
```
dannelundqvist, line 20: It is already possible to context click and choose "Open in new tab" in the context menu. So it should allow for default behaviour when using modifier key (i.e also open in new tab).

```
if (event.ctrlKey || event.metaKey) {
  return true
}
```

dannelundqvist (changes requested):
Nicely done. Really nice way of allowing new views just by creating them files in views structure. A few minor details in other comments though. Especially the NavigationState type handling and allowing modifier+click to open views in new tab.
  
Mabye picky ;) but overall I would like to see more empty lines separating logical blocks for improved readability. For example between variable declarations and function body, before each new case in switch statements and before return statements in functions. Minor things, but helps a lot when reading the code.

src/contexts/NavigationProvider.tsx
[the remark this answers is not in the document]
gularsson: Yes I struggled with this a bit and only found solutions that weren't as bad as the previous. I figured that you should be guaranteed that `state.registry` and `state.content` is always available. I guess that an initialState on creation might be the best...

src/contexts/NavigationProvider.tsx
gularsson: Yea, i figured this was temporary until we merge :slightly_smiling_face:

src/contexts/NavigationProvider.tsx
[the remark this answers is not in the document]
gularsson: By now the solution was obvious. I've separated the `init` function and are calling it in the creation of the context. Passing it to the Provider was unnecessary. I think this is even cleaner.

gularsson:
Build is failing due to changes which depends on https://github.com/ttab/elephant-ui/pull/1

dannelundqvist (approved):

PR #4, opened 2023-10-30
feat: add AppHeader and AppMenu

gularsson:


src/components/ThemeSwitcher/index.tsx
dannelundqvist, line 1: Would be nice with empty line between const declaration and return statement for readability.

src/components/ViewHeader/index.tsx
dannelundqvist, line 1: Would be nice with empty line between interface and function for readability.

src/components/index.tsx
dannelundqvist, line 1: I think it would be nice to group some of these into one folder. I.e AppMenu, AppHeader, ThemeSwitcher

dannelundqvist (changes requested):

src/components/index.tsx
gularsson: `AppHeader` is a composite view which uses, `AppMenu` and `ThemeSwitcher` are never (not now) used outside of it.

dannelundqvist (approved):
Super!

PR #5, opened 2023-10-30
Add basic editor and yjs

dannelundqvist:
Lots and lots of issues still. But basic stuff works, like logging in and opening article (hardcoded uuid). Main issue is to me that it seems too many connections are made to hocuspocus server.


src-srv/routes.ts
```diff
@@ -0,0 +1,227 @@
...
+  }
+
+  return app
+}
+
+/*
+ * Connect an exported route handler like GET, POST, etc to a specific express path.
+ */
+function connectRouteHandler(app: Application, routePath: string, func: RouteHandler, initContext: RouteInitContext): Application {
+  const handlerFunc = (req, res): void => {
```
gularsson, line 118: Could `req, res` be type as `res: Request, res: Response` instead of inferred from usage?

src-srv/routes.ts
```diff
@@ -0,0 +1,227 @@
...
+    return typeof value === 'object' && ('statusCode' in value && 'statusMessage' in value)
+  },
+  isContent: (value): value is RouteContentResponse => {
+    return typeof value === 'object' && 'payload' in value
+  }
+}
+
+/*
+ * Map directory structure under api/ to express route paths.
+ * Direcory or file names enclosed in brackets, i.e api/planning[id]/[action].ts
```
gularsson, line 62: correction => "Directory or file names enclosed in brackets, i.e api/planning/[id]/[action].ts"

src-srv/routes.ts
```diff
@@ -0,0 +1,227 @@
...
+    case 'PUT':
+      return app.put(routePath, handlerFunc)
+
+    case 'PATCH':
+      return app.patch(routePath, handlerFunc)
+
+    case 'DELETE':
+      return app.delete(routePath, handlerFunc)
+
+    // case 'WEB_SOCKET':
```
gularsson, line 166: Do we need dynamic ws-routes? Problems or how come isn't this used?

src-srv/utils/RedisCache.ts
gularsson, line 1: Terse :+1:

src-srv/utils/Repository.ts
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  /**
+   * Save a newsdoc to repository
+   * @param document Document
+   * @param documentName string
+   * @param accessToken string
+   * @returns Promise<FinishedUnaryCall<UpdateRequest, UpdateResponse>
+   */
+  async saveDoc({ document, documentName, accessToken }: { document: Doc, documentName: string, accessToken: string }): Promise<FinishedUnaryCall<UpdateRequest, UpdateResponse>> {
```
gularsson, line 138: A tad to long for my taste :slightly_smiling_face:

src-srv/utils/hocuspocus.ts
```diff
@@ -0,0 +1,64 @@
...
+  const server = Server.configure({
+    port: parseInt(process.env.API_PORT),
+    extensions: [
+      new Logger(),
+      new Redis({
+        host: process.env.REDIS_HOST, // FIXME: Should not use env
+        port: parseInt(process.env.REDIS_PORT) // FIXME: SHould not use env
+      }),
+      new Database({
+        fetch: async (data) => {
```
gularsson, line 28: I'm getting lint warnings for `async () =>`
```
  Unexpected space before function parentheses. eslint (space-before-function-paren) [28, 21]
```
Which i guess technically is correct. Dont' have a strong opinion about this.
Some more below....

src-srv/utils/hocuspocus.ts
```diff
@@ -0,0 +1,64 @@
...
+export async function createServer(options: CreateServerOptions): Promise<Hocuspocus> {
+  const { repository, cache } = options
+
+  const server = Server.configure({
+    port: parseInt(process.env.API_PORT),
+    extensions: [
+      new Logger(),
+      new Redis({
+        host: process.env.REDIS_HOST, // FIXME: Should not use env
+        port: parseInt(process.env.REDIS_PORT) // FIXME: SHould not use env
```
gularsson, line 25: Should it use the cache? Not sure how it works, if you can use the same connection for persisting of docs and pub/sub subscription?

src-srv/tsconfig.json
gularsson, line 1: Some diffs between the `tsconfig.json` for the frontend. Might be good to have the same behavior. `strict: true` for example.

src-srv/utils/transformations/index.ts
```diff
@@ -0,0 +1,48 @@
...
+
+// Add different transformers here
+const transformer = transformNewsdoc
+const reverter = revertNewsdoc
+
+export function initDoc (data: GetDocumentResponse, doc: Y.Doc): Uint8Array {
+  // TODO:Should set `meta` and `links` once a slate format i set
+  const origMap = doc.getMap('original')
+  const result = newsDocToYmap(data, origMap)
+  doc.share.set('original', result as unknown as Y.AbstractType<Y.YEvent<any>>)
```
gularsson, line 30: Not sure on how to handle the `any`. Might have to snooze the linting :man_shrugging:

src-srv/utils/transformations/newsdoc/core/text.ts
gularsson, line 1: Will have to do refactoring here, will be clearer eventually, I recognize the code and that I'm to blame.

src/views/auth/Login.tsx
```diff
@@ -0,0 +1,65 @@
...
+      <form onSubmit={(e) => { e.preventDefault() }} className="self-center flex flex-col gap-4">
+        <div>
+          <input
+            className="p-2 w-64 text-sm border border-gray-200 shadow shadow-sm rounded"
+            type="text" name="user" onChange={(e) => setUser(e.currentTarget.value)}
+          />
+        </div>
+        <div>
+          <input
+            className="p-2 w-64 text-sm border border-gray-200 shadow shadow-sm rounded"
```
gularsson, line 20: 'shadow' applies the same CSS properties as 'shadow-sm'.

src/views/auth/Login.tsx
```diff
@@ -0,0 +1,65 @@
...
+            type="password" name="password" onChange={(e) => setPassword(e.currentTarget.value)}
+          />
+        </div>
+
+        <div>
+          <button
+            className="p-2 w-64 text-sm border border-gray-300 text-bold center bg-gray-100 shadow shadow-sm rounded"
```
gularsson, line 27: 'shadow' applies the same CSS properties as 'shadow-sm'.

src/views/auth/Login.tsx
```diff
@@ -0,0 +1,65 @@
...
+      'Content-Type': 'application/json'
+    },
+    body: JSON.stringify({
+      user,
+      password
+    })
+  })
+
+  return response.status === 200
+}
```
gularsson, line 65: Could/should we use `useApi` here?

tsconfig.json
```diff
@@ -47,6 +47,7 @@
   },
   "include": [
     "src",
+    "srv-src",
```
gularsson, line 50: Shouldn't this `tsconfig.json` only be for the frontend. Not sure...

gularsson (commented):
Great refactoring, `createHocuspocusServer` and `RedisCache` are nicely modularized which creates a better entry for the `src-src/index`.ts. The `routes` approach seems very dynamic and are nicely typed, though a bit magic right now. I'll get used to it once I start using it, after all it's the Nextjs way and this way it's not all happening under the hood.

WebsocketProvider and ApiProvider :+1:

src-srv/index.ts
```diff
@@ -0,0 +1,110 @@
...
+  app.ws('/:document', (websocket, request) => {
+    hpServer.handleConnection(websocket, request)
+  })
+
+  app.listen(API_PORT)
+
+  return API_URL
+}
+
+(async () => {
```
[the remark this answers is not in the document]
dannelundqvist: Yes, when async is used in front of anon func. I have added eslint rule exception for this
("@typescript-eslint/no-confusing-void-expression": ["error", { ignoreArrowShorthand: true }])

server.ts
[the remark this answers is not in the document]
dannelundqvist: Yes!

src-srv/api/documents/index.ts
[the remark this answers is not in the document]
dannelundqvist: Yes, the file is not used, see it as a template as we don't have any other documentation atm.

src-srv/api/ws.ts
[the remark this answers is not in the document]
dannelundqvist: ...

src-srv/api/documents/[id]/index.ts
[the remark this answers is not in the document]
Comment (dannelundqvist) [reply src-srv/api/documents/[id]/index.ts:1]: Yes, the file is not used, see it as a template as we don't have any other documentation atm.

src-srv/index.ts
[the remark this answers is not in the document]
dannelundqvist: We can rename it. The simple idea is to have src- as prefix so it sits nicely next to the src directory.

src-srv/routes.ts
dannelundqvist: Good catch!

src-srv/routes.ts
dannelundqvist: Leftover comment, I'll remove it. Websockets are handled in another function below as it works differently.

tsconfig.json
dannelundqvist: Not sure either... let's look into it together

src/views/auth/Login.tsx
dannelundqvist: I have refactored api/session almost completely now.

src-srv/utils/transformations/newsdoc/core/text.ts
dannelundqvist: Yes, I think we'll have to rewrite transformations completely,

src-srv/utils/transformations/index.ts
dannelundqvist: We'll look into the transformations next. Ignoring it for now ;)

src-srv/utils/hocuspocus.ts
dannelundqvist: Yes, when async is used in front of anon fat arrow func. I have added eslint rule exception for this
("@typescript-eslint/no-confusing-void-expression": ["error", { ignoreArrowShorthand: true }])

src-srv/utils/hocuspocus.ts
dannelundqvist: I think it is needed. It might be that the code for the Database extension is wrong. It should probably not fetch from redis as it is doing now. Let's investigate it together

gularsson: For documentation: I think this is needed with multiple instances, not needed if the instance already have a cached version of the document. But if we're running multiple instances and the current instance haven't served it previously we 
need to fetch the complete document from redis. 

But there's probably room for improvement.

PR #6, opened 2023-11-02
feat: add avatar

gularsson:
Depending upon: https://github.com/ttab/elephant-ui/pull/6

### DIFF
--- a/src/types/index.ts
+++ b/src/types/index.ts
@@ -51,6 +51,8 @@ export interface HistoryState {
 
 export interface ViewProps {
   id: string
+  name: string
+  index?: number
 }
 
 export type Theme = 'dark' | 'light' | 'system'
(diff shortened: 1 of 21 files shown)


src/types/index.ts
dannelundqvist, line 1: I am not 100% sure on the jwt interface name. This will live in the Elephant only, would it be so crazy to skip the "Elephant" prefix? (If you have strong opinions I'm not going to argue though.) Also, as it is an acronym it might be good to keep JWT in all capitals.

dannelundqvist (approved):
Like how the UI components are coming together now. Good work!

src/types/index.ts
gularsson: I agree, Elephant as a prefix is unnecessary, there might be a point in indicating that its not the original interface but an extended though. One that suits our purposes.

`JWT` is pretty clear though, I'll use that instead.

PR #7, opened 2023-11-03
feat: add cmdk navigation

gularsson:


src/components/Link/link.ts
gularsson, line 1: Not sure about the naming here. Because of `react-refresh` we need to separate component and utility functions to different files.

dannelundqvist: I think it is ok, and a good practice to have util funcs next to the component like this.

dannelundqvist (commented):
Works really nice when testing it!

src/components/CommandMenu/index.tsx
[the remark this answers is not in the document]
gularsson: Guess we could add a modifer check in `handleLink` and if we have a selected action with modifier we'll call `window.open` instead

dannelundqvist (approved):
Nice work!

PR #8, opened 2023-11-06
Feature/ele 179 make hocuspocus conn in client work with multiple

dannelundqvist:
Pulling in auto refresh token, refactored auth flow, session handling, api handling and added multiplexing support by sharing websocket for all hocuspocus instances in client.


src/components/Link/index.tsx
```diff
@@ -35,8 +39,7 @@ export const Link = ({ children, to, props }: LinkProps): JSX.Element => {
         ...history.state.contentState,
         {
           id,
-          name: linkItem.metadata.name,
-          props
```
gularsson, line 39: I think we need to keep `props` here...
This is for when we with back navigation need to recreate a previous, now unmounted, component.

src/contexts/ApiProvider.tsx
```diff
@@ -1,80 +1,38 @@
...
+  apiUrl: string
+  websocketUrl: string
 }
 
 export interface ApiProviderState {
-  api?: string
-  ws?: string
-  jwt?: JWTPayload
+  apiUrl: string
+  websocketUrl: string
```
gularsson, line 12: Would be nice to treat these as an `URL`. Might not be possible but it would simplify some of the handling, appending, validation...

gularsson (approved):

src/components/Link/index.tsx
gularsson: It's possible that we need to revisit this later on, right now it _might_ work without, but when we need to persist state for unmounted views, I think this is the place to do it with the current solution.

dannelundqvist: I'll leave it for later then.

src/contexts/ApiProvider.tsx
dannelundqvist: Agree. I'll convert it.

PR #11, opened 2023-11-08
chore: snooze lint, fix tests

gularsson:


src-srv/utils/transformations/index.ts
dannelundqvist, line 1: The recommended way would be to use unknown and cast it multiple times (!) — but — here "any" is defined by yjs types. So not sure we can do this without rewriting yjs... I suggest we skip the "FIXME:", leave a comment on why, and move on.

src-srv/utils/transformations/yjs/index.ts
dannelundqvist, line 1: Same as above. The recommended way would be to use unknown and cast it multiple times (!) — but — here "any" is defined by yjs types. So not sure we can do this without rewriting yjs... I suggest we skip the "FIXME:", leave a comment on why, and move on.

dannelundqvist (changes requested):
Nice to have it cleaned up!

dannelundqvist (approved):

PR #12, opened 2023-11-08
Feature/ele 180 focus action should live in each view header

gularsson:


src/components/NavigationWrapper/index.tsx
dannelundqvist, line 1: Instead of `pl-2` and `first:border-none` we should use `gap` and let flex handle this.

dannelundqvist (changes requested):
Nicely done. Just would like to see a tiny css simplification. Then I think we should discuss `lib/init.tsx` outside of this PR though.

dannelundqvist (approved):

PR #15, opened 2023-11-14
Feature/ele 191 planning view

gularsson:
Needs https://github.com/ttab/elephant-ui/pull/10


src/components/PlanningTable/RowActions.tsx
```diff
@@ -0,0 +1,51 @@
...
+        <Button
+          variant="ghost"
+          className="flex h-8 w-8 p-0 data-[state=open]:bg-muted"
+        >
+          <DotsHorizontalIcon className="h-4 w-4" />
+          <span className="sr-only">Open menu</span>
+        </Button>
+      </DropdownMenuTrigger>
+      <DropdownMenuContent align="end" className="w-[160px]">
+        {row.original._source?.['document.meta.core_assignment.rel.deliverable.uuid']
```
dannelundqvist, line 36: An empty array of deliverable uuids would enter the loop and output nothing where I would probably expect 'No deliverables' (if that was the intention). In addition, a ternary with long property checks and inline loops inside of rendering makes it difficult to read and understand. Could (or should...) be rewritten as something like:

```
const deliverableUuids = row.original._source?.['document.meta.core_assignment.rel.deliverable.uuid'] || []

return (
  ...
  <DropdownMenuContent align="end" className="w-[160px]">
    {!deliverableUuids.length && 'No deliverables}
    {deliverableUuids.length && deliverableUuids.map(uuid => {
       return <DropdownMenuItem key={uuid>
         <Link ...>
           {uuid}
         </Link>
       </DropdownMenuItem>
    }
  </DropdownMenuContent>
   ...
)
```

src/components/PlanningTable/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+      <DataTableToolbar table={table} />
+      <div className="rounded-md border">
+        <Table>
+          <TableHeader>
+            {table.getHeaderGroups().map((headerGroup) => (
+              <TableRow key={headerGroup.id}>
+                {headerGroup.headers.map((header) => {
+                  return (
+                    <TableHead key={header.id}>
+                      {header.isPlaceholder
```
dannelundqvist, line 81: Would be easier to read as a oneline, even if a tad bit long.

```
{header.isPlaceholder ? null : flexRender(header.column.columnDef.header, header.getContext())}
```

dannelundqvist: A little bit picky ;)

src/components/PlanningTable/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+                          header.getContext()
+                        )}
+                    </TableHead>
+                  )
+                })}
+              </TableRow>
+            ))}
+          </TableHeader>
+          <TableBody>
+            {table.getRowModel().rows?.length
```
dannelundqvist, line 94: Large ternaries makes it difficult to follow the logic, Should be refactored into it's own helper render functions in the same file.

dannelundqvist (changes requested):
Lots of work done. Great job with the typings! Most comments are about ternaries making the rendering logic hard to follow in many cases.

src/components/PlanningTable/ColumnHeader.tsx
```diff
@@ -0,0 +1,75 @@
...
+    <div className={cn('flex items-center space-x-2', className)}>
+      <DropdownMenu>
+        <DropdownMenuTrigger asChild>
+          <Button
+            variant="ghost"
+            size="sm"
+            className="-ml-3 h-8 data-[state=open]:bg-accent"
+          >
+            <span>{title}</span>
+            {column.getIsSorted() === 'desc'
```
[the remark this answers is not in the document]
gularsson: Great, will use this :+1:

src/components/PlanningTable/Pagination.tsx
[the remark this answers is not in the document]
gularsson: Good idea, now it's tied to react-table functions, but you could probably make i generic for search params and/or generic react-table.

src/components/PlanningTable/index.tsx
gularsson: I'll think I'll leave this, it requires a change/ignore of eslint rules to be able to keep it on one line.

src/components/PlanningTable/RowActions.tsx
gularsson: Good catch!

PR #18, opened 2023-11-22
Feature/ele 212 focused view should be reflected in browser address and ux

gularsson:
# Changes to the navigation experience
Current active view is being persisted and updated in NavigationProvider as `state.active`. A border is added to the top of the active window to indicate active state. Active is changed by clicking in a view. When a user interaction opens a new view, it should become the active one.
Todo: cmdk change active view, keyboard navigation between views

`useHistory` is augmented to track what type of history action is being made.

`popstate` - regular user navigation with browsers back/forward and initial load
`pushstate` - user interaction opens a new view
`replacestate` - url rewrites

```
const historyState = useHistory()
historyState.type => popstate/pushstate/replacestate
```

With the above types declared we can rewrite urls when `state.active` changes with out rerendering (:pray:), we can better see what navigation action should be taken and overall reduce unnecessary rerenders.

Navigation components have been moved to `src/navigation/`.


dannelundqvist (approved):
A lot easier to dig into the navigation code now with this file and directory restructure. Really good work! I am a bit afraid the onClick handler in NavigationWrapper might not work in all situations but I think we should look at it again when more UI have been created. Not meaningful to try and guess future issues :)

PR #19, opened 2023-11-24
fix: hide ViewFocus if one view, fix id

gularsson:


dannelundqvist (changes requested):
If you updated to textbit 0.5.0 you get correct exports for TextbitEditable and make the checks pass.

dannelundqvist (approved):

PR #20, opened 2023-11-24
Feature/ele 221 styles in textbit should use elephant chrome styles

dannelundqvist:
A somewhat misnamed branch as using the variables from chrome-ui was no real issue. Updated to use all renamed Textbit types correctly. Removed unnecessary dep to slate-yjs.


gularsson (approved):
Great :+1:

PR #24, opened 2023-11-28
test: mock planning search data

gularsson:


dannelundqvist (approved):
+1 We need more tests :)

PR #25, opened 2023-11-28
feat: remove view that dont fit

gularsson:
Partial of https://linear.app/tt-nyhetsbyran/issue/ELE-183/views-in-the-app-should-be-addedremoved-when-window-size-changes

In this:
* Add `useResize` to be able to register resize
* Remove if view doesn't fit new size

To be able to add a view after resize to bigger viewport we need to have a discussion about how. Might be a weird pattern if we readd views that have been removed?
And we need to set a overall strategy for `min-width` and `max-width` for views.


dannelundqvist (approved):
Nice schoolbook example of useSyncExternalStore!

PR #27, opened 2023-11-29
Feature/ele 228 update link component structure

dannelundqvist:
According to lates discussion about code structure


gularsson (approved):
:+1:

PR #30, opened 2023-12-04
feat: refine planning-list

gularsson:


src/components/PlanningTable/Columns.tsx
dannelundqvist, line 1: This file has grown quite bit. Though well structured it is a lot. Would it be an idea to try and split it into one main Columns.tsx and then a column component for each column type and have that in it's own directory? If possible it will probably be easier to see the structure, change one column type in isolation as well as adding more "column types" in the future.

dannelundqvist (changes requested):
Nice work! Just one comment on the Columns component which is growing big.

src/components/PlanningTable/Columns.tsx
gularsson: Yes, sounds reasonable. Could we postphone the restructuring to another PR? Now there is three PRs that together refines the planning/header. I think it would be easier to do it once all the pieces have been put together.

dannelundqvist: Definitely!

dannelundqvist (approved):

PR #32, opened 2023-12-06
Feature/ele 251 refactor columnstsx

gularsson:


src/components/PlanningTable/Columns/Title.tsx
```diff
@@ -0,0 +1,38 @@
...
+              </TooltipTrigger>
+              <TooltipContent>
+                <p>{status ? 'Public' : 'Internal'}</p>
+              </TooltipContent>
+            </Tooltip>
+          </TooltipProvider>
+          <span className='max-w-[200px] md:max-w-[300px] lg:max-w-[700px] truncate font-medium'>
+            {row.getValue('title')}
+          </span>
+          {slugline?.length && (
```
dannelundqvist, line 32: This type of construct will happily render a "0" as React does not turn it into a boolean. It´s safer to always make sure it is a boolean like
`{!!slugline?.length && ( ...)}`

src/components/PlanningTable/Columns/Title.tsx
```diff
@@ -0,0 +1,38 @@
...
+  cell: ({ row }) => {
+    const status = row.original._source['document.meta.core_planning_item.data.public'][0] === 'true'
+    const slugline = row.original._source['document.meta.core_assignment.meta.tt_slugline.value']
+
+    return (
+        <div className='flex space-x-2 w-fit'>
+          <TooltipProvider>
+            <Tooltip>
+              <TooltipTrigger>
+                <div className={cn(
```
dannelundqvist, line 18: The code would be easier to read if we moved the `cn()` call outside of the JSX and store it as variable. Later on we should use `cva()` from `class-variance-authority` for similar class name conditionals.

dannelundqvist (changes requested):
Huge improvement! Easy to find specific columns cell implementation now. Will be easy to change and add/remove column cell types.

Only found one small issue with a conditional rendering (and a mandatory nitpicky code structure thingy).

dannelundqvist (approved):

PR #34, opened 2023-12-07
fix: use resolved preset for settings

gularsson:
Waiting for: https://github.com/ttab/elephant-ui/pull/14


dannelundqvist (approved):
As said in elephant-ui comment. Very good solution!

PR #37, opened 2023-12-11
feat: build server

gularsson:
depends on: https://github.com/ttab/elephant-ui/pull/16


dannelundqvist (changes requested):
Some really good QA-fixes and cleanups. Good work in getting the build work well!

I can't see the benefit of removing the possibility to change port for the client. It's good to have a fallback but also to be able to control it. Unless there are compelling reasons not to, I'd love to see that env variable back.

gularsson:
> Some really good QA-fixes and cleanups. Good work in getting the build work well!
> 
> I can't see the benefit of removing the possibility to change port for the client. It's good to have a fallback but also to be able to control it. Unless there are compelling reasons not to, I'd love to see that env variable back.

The previous setup was kinda dev-centric. The client only need a port when we're running the "devserver" for local development. In production the bundle are served from express, hence we only have a port for our backend. I think it should be a prod-first approach, ie no code for beeing able to run the devserver, instead I'm using the vite server.proxy to simulate a more prod-like way of serving it locally. (I think there might be more things that we can remove)

dannelundqvist (approved):

PR #42, opened 2023-12-13
feat: use isSynced to determine connection to doc

gularsson:


dannelundqvist (approved):
Looks good! I also think it is good to keep it simple for now.

PR #45, opened 2023-12-13
fix: reserv width for AppHeader, transparent-border for non-actives

gularsson:


dannelundqvist (changes requested):
Probably not related to these changes but the app header does not align horizontally with the view header. Either way, it would be nice to fix. Other than that it's great!

gularsson:
> Probably not related to these changes but the app header does not align horizontally with the view header. Either way, it would be nice to fix. Other than that it's great!

Did with two views, but not one, header width varies 57px (aligns) 53px(doesn't align). I'll fixate that tomorrow.

gularsson:
I set a fixed height for the ViewHeader so it wont resize depending on it's content. Should keep the borders aligned.

dannelundqvist (approved):
Perf!

PR #46, opened 2023-12-13
Feature/ele 273 keyboard navigation in planningoverview

dannelundqvist:
Added initial boilerplate code for navigation up/down in planning list using arrow keys. I removed support for multi line selections to simplify the code. If no callback is provided it is not possible to select rows as it does not make sense to select rows without functionality connected to it.

Would it make more sense to send just the data id back to the callback?

### DIFF
--- a/src/components/PlanningTable/index.tsx
+++ b/src/components/PlanningTable/index.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react'
+import React, { useEffect, useState } from 'react'
 import {
   type ColumnDef,
   type ColumnFiltersState,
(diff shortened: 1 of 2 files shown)


src/components/PlanningTable/index.tsx
```diff
@@ -14,30 +14,22 @@ import {
...
-  data
-}: PlanningTableProps<TData, TValue>): JSX.Element => {
-  const [rowSelection, setRowSelection] = React.useState({})
-  const [columnVisibility, setColumnVisibility] =
-    React.useState<VisibilityState>({})
-  const [columnFilters, setColumnFilters] = React.useState<ColumnFiltersState>(
-    []
-  )
-  const [sorting, setSorting] = React.useState<SortingState>([])
+export const PlanningTable = <TData, TValue>({ columns, data, onRowSelected }: PlanningTableProps<TData, TValue>): JSX.Element => {
```
gularsson, line 28: I suppose this now a one liner for readability, and with that I some what agree. Unfortunately I think this decreases the visibility, in my opinion 80-100 chars is the max length for a line.

src/components/PlanningTable/index.tsx
```diff
@@ -61,6 +54,55 @@ export const PlanningTable = <TData, TValue>({
...
+      if (evt.key === 'Escape') {
+        if (selectedRow) {
+          selectedRow.toggleSelected(false)
+        }
+      } else if (!selectedRow) {
+        const idx = evt.key === 'ArrowDown' ? 0 : rows.length - 1
+        rows[idx].toggleSelected(true)
+      } else {
+        const nextIdx = selectedRow.index + ((evt.key === 'ArrowDown') ? 1 : -1)
+        const idx = nextIdx < 0 ? rows.length - 1 : nextIdx >= rows.length ? 0 : nextIdx
```
gularsson, line 86: Maybe this could be a function? Nested ternaries are a bit hard to read.
Maybe do something about if-if-else-if-else-chunk as well?

gularsson (approved):
Great add to the overall experience and a good start to be able to take actions on rows :star_struck:

src/components/PlanningTable/index.tsx
dannelundqvist: Agree that it is not perfect. Will leave it unchanged as we will refactor this when we have refactored NavigationWrapper into ViewWrapper and then add hooks to simplify this.

PR #47, opened 2023-12-14
feat: add time column

gularsson:


src/lib/getPublishTime.ts
dannelundqvist, line 1: As all dates/times in repo/opensearch are in UTC, and we always need to display in local time, it is better to convert to (generic) Date() object immediately as this will also take timeZone into account. Then it is easier to always just convert it to displayable local format when needed. This principle will give us fewer issues.

I would like to se
```
export function getPublishTime(assignmentPublishTimes: string[]): Date | undefined {
  ...
  const date = new Date(isoString)
  if (isNaN(date.getTime()) || date.toString() === 'Invalid Date') {
    return
  }
  ...
}
```

src/lib/datetime.ts
dannelundqvist, line 1: I think we should always require the calling party to provide a valid Date object, locale and timeZone as it will always be needed.
  
We also need to take into account that some locales use 12 hour clock and am/pm. We could add a function to datetime.ts like below:

```
function isAMPMPreferred(locale) {
  const date = new Date()
  const options = {
    hour12: true, // Use 12-hour clock (AM/PM)
  }

  try {
    const dateTimeFormatter = new Intl.DateTimeFormat(locale, options)
    const formattedDate = dateTimeFormatter.format(date)
    return formattedDate.includes('AM') || formattedDate.includes('PM')
  } catch (error) {
    return false // Default false on errors
  }
}
```

Then the code in the original function could be:
```
  return new Intl.DateTimeFormat(locale, {
    hour12: isAMPMPreferred(locale),
    hour: '2-digit',
    minute: '2-digit',
    timeZone
  }).format(date)
  ```

Though whether am/pm is preferred should probably be put into the Registry alongside locale and timeZone (and added as parameter to this function) so it is not calculated every time we display a time.

And with all these changes the original function should probably be renamed as well ;)

**Review by dannelundqvist — CHANGES_REQUESTED** [review]
Good thinking in breaking things up. We need to handle date/time a consistently and I have added a suggestion on how to make this more generic and future proof.

File: src/views/PlanningOverview/PlanningGridColumn.tsx
Comment (dannelundqvist) [inline src/views/PlanningOverview/PlanningGridColumn.tsx:1]: Sorry for being obnoxious here ;) but now that you've made these changes this can be a lot simpler. Function `getPublishTime()` returns  either valid date or undefined... So you can do this:

```
const assignmentDataPublish = getPublishTime(item._source['document.meta.core_assignment.data.publish'])

// Unnecessary lines...
// const startTime = assignmentDataPublish
//            ? dateToReadableTime(assignmentDataPublish, locale, timeZone)
//            : null

{!!assignmentDataPublish &&
  <div className="flex-none text-muted-foreground">
    {dateToReadableTime(assignmentDataPublish, locale, timeZone)}
  </div>
}

```

dannelundqvist (changes requested):

src/views/PlanningOverview/PlanningGridColumn.tsx
gularsson: Good catch, will fix

PR #49, opened 2023-12-15
Feature/ele 289 use base for serving app

gularsson:
I think this will do.
I've simplified the envs. If we use vite's server.proxy we don't need to care about API_HOSTS and API_PORTS. This is modeled to be as the prod server is running, express serving the bundle.

:crossed_fingers:

### DIFF
--- a/__tests__/commandMenu.test.tsx
+++ b/__tests__/commandMenu.test.tsx
@@ -22,6 +22,6 @@ describe('CommandMenu', () => {
     await userEvent.keyboard('{Control>}k')
     expect(screen.getByText('Planning overview')).toBeInTheDocument()
     await userEvent.click(screen.getByRole('option'))
-    expect(history.state.itemName).toBe('PlanningOverview')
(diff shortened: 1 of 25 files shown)


dannelundqvist (changes requested):
Very nice refactor and cleanup of unnecessary endpoint juggling. One thing I am a bit worried about is the use of PascalCase in paths (PlanningOverview). It might be a good idea if we start using lowercase only so we don't have to juggle case in currentView etc. It feels like something that could bite us later on.

gularsson:
> Very nice refactor and cleanup of unnecessary endpoint juggling. One thing I am a bit worried about is the use of PascalCase in paths (PlanningOverview). It might be a good idea if we start using lowercase only so we don't have to juggle case in currentView etc. It feels like something that could bite us later on.

Yes, there is a point in keeping it more simple. This is unnecessary juggling.

dannelundqvist (approved):

PR #53, opened 2023-12-19
fix: serve init urls from api

gularsson:


src/contexts/ApiProvider.tsx
dannelundqvist, line 1: I think it would be good if we check the response and could have an ApiProvider state variable that indicates whether api is initalized (or similar).

```javascript
const response = await fetch('/api/init')
if (response.ok) {
   const urls = await response.json()
   seturls(...)
   setApiInitialized(true)
 }
 ```
 
 Then we would be able to show message, or do something, when we don't have the right urls. If needed etc etc.

dannelundqvist (changes requested):
A lot cleaner than exposing them directly in the client code. Only a small response validation request from my part.

dannelundqvist (approved):

PR #55, opened 2023-12-20
fix: remove colors from assignee avatars

gularsson:


dannelundqvist (approved):
Nice. No more comments.

PR #58, opened 2023-12-22
Feature/ele 281 enable keyboard navigation in planning list view using

dannelundqvist:
Added date selection with left/right arrow keys. Ignore arrow key navigation when editable fields are event targets (input, select, isContentEditable etc.


gularsson (approved):
Looks good :+1:

PR #62, opened 2023-12-22
Feature/ele 263 add container breakpoints for planning list so that it

dannelundqvist:
Added @container style handling so it is possible to adjust children to actual view width breakpoints.


gularsson (approved):
👍

PR #63, opened 2024-01-03
fix: hide other views, add scrollarea, reset focus on nav

gularsson:
fixes: 
https://linear.app/tt-nyhetsbyran/issue/ELE-300/other-views-are-viewable-in-the-background-when-in-focused-mode
https://linear.app/tt-nyhetsbyran/issue/ELE-206/add-scroll-area-to-navigationwrapper
https://linear.app/tt-nyhetsbyran/issue/ELE-326/sort-plannings-desc
https://linear.app/tt-nyhetsbyran/issue/ELE-325/sticky-header

extra reset focus on navigation

depends on: https://github.com/ttab/elephant-ui/pull/20


dannelundqvist (changes requested):
Made a huge difference in how the app feels. Fantastic!

As per discussion:
1. The flake files should not be in the repo.
2. Renaming `hasFocused` to `isHidden` would make it more self explanatory.

dannelundqvist (approved):
Nice!

PR #64, opened 2024-01-03
fix: overlapping elements

gularsson:
I'm unable to find a good way to specify column widths, this pr solves a ui bug, but we still need to figure out how to tighten up the columns. Especially the static content ones with fixed widths.


dannelundqvist (approved):
No comments :)

PR #65, opened 2024-01-04
Feature/ele 324 update textbit in elephant chrome to new structure

dannelundqvist:
* Updated Textbit to version 0.6.5 and use hook to create independent editor footer in elephant-chrome instead of using default.
* Moved scrolling behaviour to each view instead of general View for improved ux
* Removed overscroll bounce on app (body)


index.html
```diff
@@ -1,14 +1,17 @@
...
+  <link rel="icon" type="image/svg+xml" href="/vite.svg" />
+  <link href="/dist/output.css" rel="stylesheet">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+  <title>Elephant</title>
+</head>
+
+<body style="overscroll-behavior: none; height: 100vh; overflow: hidden;">
+  <div id="root"></div>
+  <script type="module" src="/src/main.tsx"></script>
+</body>
```
gularsson, line 15: Seems like something happened with the indentation here?

src/views/Editor/index.tsx
```diff
@@ -88,18 +92,46 @@ const Editor = (props: ViewProps): JSX.Element => {
...
+        <span>{words}</span>
+      </div>
+      <div className="flex gap-2">
+        <strong>Characters:</strong>
+        <span>{characters}</span>
+      </div>
+    </div>
+  )
+}
+
```
gularsson, line 134: Maybe use `<footer>` to keep the semantics?

src/views/Editor/index.tsx
```diff
@@ -88,18 +92,46 @@ const Editor = (props: ViewProps): JSX.Element => {
...
+              <TextbitEditable yjsEditor={editor} />
+            </div>
+          </ScrollArea>
+
+          <div className="grow-0 border-t opacity-90">
+            <Footer />
+          </div>
         </div>
-      </main>
+
```
gularsson, line 112: Maybe keep `<main />` here as well, or is that superfluous with `<Textbit />`?

gularsson (changes requested):
Great, view based scrolling was a better solution than a more generic one. Looks great.

Not sure if it's just on my system but the vertical scroll is only half visible (firefox/chromium). But since vertical scroll isn't desirable and should be solved with better responsivness, i think we can leave it like this.

src/views/Editor/index.tsx
dannelundqvist: If we want to be picky we should probably not use `<main>` the way we do as we have several views open at the same time. The spec states:

*A document must not have more than one [main](https://html.spec.whatwg.org/multipage/grouping-content.html#the-main-element) element that does not have the [hidden](https://html.spec.whatwg.org/multipage/interaction.html#attr-hidden) attribute specified.*

And as the _hidden_ attribute tells the browser that the element should not be rendered we should not used that.

I think it is better we lean on having `<section>` in the View component.

gularsson: Yes, that sounds correct

gularsson (approved):

PR #66, opened 2024-01-09
Feature/ele 271 filter menu planningoverview

gularsson:
Depends on: https://github.com/ttab/elephant-ui/pull/21

First stab, there's room for improvement. We need to refine the cmd-k experience a bit, add shortcuts and generalize.

There's also some optimization to be done, table loading is (sometimes) slow, I've left that for now 'cause I think that's an overall problem for the table at the moment.


src/components/DataItem/SectorBadge.tsx
dannelundqvist, line 1: Generic components should not have knowledge about more specific usage files. But we'll fix this when moving the data part instead later :)

src/views/PlanningOverview/PlanningTable/Columns/Section.tsx
dannelundqvist, line 1: It is beneficial to keep naming consistent to avoid confusion. I would suggest we rename this to Sector as that is the name of the data it displays.

dannelundqvist (changes requested):
Impressive work! I really like how the command+k dialog is coming along.

Would it be possible to "clear the state" of the command+k-dialog when it is closed? It is confusing when you reopen it and it might still be deep into a specific filter option.

Other than that and specific compnents I think the solution is really nice!

PR #67, opened 2024-01-12
Feature/ele 335 poc concurrent editing of news value or other metadata

dannelundqvist:


src/views/Editor/Toolbar.tsx
gularsson, line 1: Shouldn't this be in `/Toolbar` dir?

src/views/Editor/Toolbar.tsx
```diff
@@ -0,0 +1,63 @@
...
+    isSynced,
+    document,
+    initNewsvalueDuration,
+    initNewsvalueScore,
+    initNewsvalueEnd
+  ])
+
+  return (
+    <>
+      <NewsValueScoreDropDown
```
gularsson, line 39: Maybe add Tooltip for NewsValueScore and NewsvalueTime?

src/views/Editor/Toolbar/NewsValueTimeDropDown.tsx
```diff
@@ -0,0 +1,172 @@
...
+      <PopoverTrigger asChild>
+        <Button
+          variant="ghost"
+          className="flex h-8 p-0 px-2 mb-1 data-[state=open]:bg-muted"
+        >
+          <span className="flex items-end">{isEnd ? dateToReadableDateTime(new Date(end), locale, timeZone) : value}</span>
+        </Button>
+      </PopoverTrigger>
+
+      <PopoverContent className="w-[320px]">
```
gularsson, line 47: Since the default alignment for Popover is center, which looks best imo, the content drifts when switching between `∞`, `24h` `<datetime>`. Maybe using `align="start"` might be better?

gularsson (changes requested):
Great! Good to have proven that editing multiple parts of a document is possible, and just certain values. The `useYmap` hook looks great.
This should make it possible to move forward with planning, and especially making changes directly from the planning row.

I mostly raise questions that are of a more structural character.

src-srv/utils/transformations/index.ts
```diff
@@ -20,23 +18,30 @@ export interface SlateDoc {
...
-}
 
-export function revert(data: SlateDoc): GetDocumentResponse {
+/**
+ * Convert Y.Doc to Slate document
+ */
+export function toNewsDoc(data: SlateDoc): GetDocumentResponse {
   if (data.document !== undefined) {
     return {
       ...data,
```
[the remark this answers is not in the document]
dannelundqvist: 👍

src-srv/utils/transformations/index.ts
```diff
@@ -20,23 +18,30 @@ export interface SlateDoc {
...
-  doc.share.set('original', result as unknown as Y.AbstractType<Y.YEvent<any>>)
+//   // Set original
+//   const yMap = newsDocToYmap(data, yDoc.getMap('original'))
+//   // eslint-disable-next-line @typescript-eslint/no-explicit-any
+//   yDoc.share.set('original', yMap as unknown as Y.AbstractType<Y.YEvent<any>>)
+
+//   // Format and add content as slate editable
+//   const content = transformer(data.document?.content ?? [])
+//   return yjsStateAsUpdate(content, yDoc)
+// }
```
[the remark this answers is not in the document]
dannelundqvist: Your right! I did a real cleanup of this now. To follow your suggestions and how we do it elsewhere.

src-srv/utils/CollaborationServer.ts
```diff
@@ -0,0 +1,212 @@
...
+      return state
+    }
+
+    // Fetch content
+    const documentResponse = await this.#repository.getDoc({
+      uuid,
+      accessToken: context.token
+    })
+    const { document } = documentResponse
+    console.log(JSON.stringify(document, null, 2))
```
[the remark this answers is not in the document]
dannelundqvist: 🧐

src/hooks/useYjsMap.tsx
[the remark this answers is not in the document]
dannelundqvist: First I didn't like this idea as it only handles a yMap structure. But then again, if we expand on it we will probably use the yMap structure as the base anyways. 👍

src/views/Editor/Toolbar.tsx
dannelundqvist: Yes, and even more, I renamed it to EditorHeader to follow the same pattern as for Planning view.

src/views/Editor/Toolbar/NewsValueTimeDropDown.tsx
dannelundqvist: Perfect!

src/views/Editor/Toolbar.tsx
dannelundqvist: 👍

gularsson (approved):

PR #68, opened 2024-01-15
fix: separate columns and components, useMemo

gularsson:


dannelundqvist (approved):
Very nice cleanup, better structure.

PR #70, opened 2024-01-17
fix: table cells container queries

gularsson:
I was unable to get the builtin column settings `size`, `minSize` and `maxSize` to work at all. There seems to be a lot of confusion on how to actually set column size in `react-table v8`.

I use the `column.columnDef.meta` to add a `className` key, with which we'll decorate each column. Together with `table-fixed` we can set fixed sizes for all columns. `Title` is kept without a fixed width.

Now we're hiding columns when screen/container width decreases, some of these we might want to present in other ways rather than hiding, others we might want to hide inside the Actions menu.


src/views/PlanningOverview/PlanningTable/Columns/Sector.tsx
dannelundqvist, line 1: Duplicate data. Maybe not the purpose of this PR/ticket, but the sector data should be fetched from `@defaults/sectors.tsx`. As it is now the label has been changed here but not elsewhere.

dannelundqvist (approved):

PR #107, opened 2024-01-22
Feature/ele 355 planningview show assignments

gularsson:


src/views/Planning/components/PlanDate.tsx
```diff
@@ -0,0 +1,63 @@
...
+  return (
+    <div className='flex w-fit space-x-2'>
+      <Calendar className='h-4 w-4' color='#4848FA' />
+      <span className='text-sm font-normal leading-4'>{formatDate(start as string, locale, timeZone)}</span>
+      {start !== end && <span className='text-sm font-normal leading-4'>{formatDate(end as string, locale, timeZone)}</span>}
+    </div>
+  )
+}
+
+function formatDate(value: string | undefined, locale: string, timeZone: string): string | null {
```
dannelundqvist, line 37: The formating part should probably live well in _@lib/datetime.ts_ and be renamed to dateToReadableDate(). It could also be improved (both less error prone and better showing intent) by letting Date parse the date and using builtin functions to manipulate date.

```
const date = new Date(foo)
foo.setHours(0, 0, 0, 0)
foo.setDate(foo.getDate() - 1)
```

dannelundqvist (changes requested):
Nice work! I found a potential issue with date handling, see comment. Otherwise, just nice 👍

src/views/Planning/components/PlanDate.tsx
gularsson: Yes, I agree. There was some questions regarding what type of data we were receiving. But it seems like we only get a string `yyyy-mm-dd`, same value `start` as `end`.
So i kinda hacked it together.

I think the question has been cleared and i'll make a more persistant solution.

gularsson: After discussion: I'll revert and leave it as is. `start` and `end` are, as of now, provided as a string with no indication of timezone.

dannelundqvist (approved):

PR #113, opened 2024-01-23
Feature/ele 365 remote collaboration users should be displayed in app header

dannelundqvist:
Added awareness for "documents" and displaying avatars for remote users that have an article open. Still some issues around avatars appearing/disappearing but enough for demo purposes. Merged conflicting generalization of Avatar component into one (added more generic view agnostic variants and added support for remote user colors).


src/components/AvatarGroup.tsx
```diff
@@ -0,0 +1,11 @@
+import { type PropsWithChildren } from 'react'
+
+interface AvatarGroupProps extends PropsWithChildren {}
```
gularsson, line 3: Isn't PropsWithChildren enough as interface for the component?

src/components/AvatarGroup.tsx
gularsson, line 1: Could this be used in AssigneeAvatar as well? :thinking:

src/components/Awareness.tsx
```diff
@@ -7,20 +7,31 @@ import { useAwareness } from '@/hooks'
...
   })
 
-  const color = Collaboration.colors[focused?.color || ''] || ''
+  // Currently we only show one user color as ring around remotely focused element
+  const state = states.length ? states[0] : null
+  const color = Collaboration.colors[state?.focus?.color || '']?.ring || ''
+
   return (
-    <div ref={awarenessRef} className={`rounded ${focused ? 'ring' : ''} ${color}`}>
+    <div ref={awarenessRef} className={`rounded ${visual && !!color ? 'ring' : ''} ${color}`}>
```
gularsson, line 34: Feels like this could be clearer. `state` isn't typed as possible undefined, yet we check for length.
Maybe use `cn()` instead of ternary?

src/components/ViewHeader/ViewHeader.tsx
```diff
@@ -33,11 +38,43 @@ export const ViewHeader = ({ children, title, shortTitle, icon: Icon }: ViewHead
...
         }
       </div>
     </header >
   )
 }
+
+function RemoteUsers({ documentId }: PropsWithChildren & { documentId: string }): JSX.Element {
+  const [states] = useAwareness(documentId)
+
+  const users = states.reduce<Array<{ clientId: number, data: AwarenessUserData }>>((remoteUsers, state) => {
```
gularsson, line 55:
```suggestion
  const users = states.reduce<AwarenessStates>((remoteUsers, state) => {
```

src/views/Planning/components/PlanDescription.tsx
```diff
@@ -24,27 +24,24 @@ export const PlanDescription = ({ isSynced, document }: CollabComponentProps): J
...
+          if (setFocused.current) {
+            console.log(event)
+            setFocused.current(false)
+          }
+        }}
+        className='h-screen'
       />
-      </Awareness>
-      )
+    </Awareness>
```
gularsson, line 45: Would prefer the previous indentation?

gularsson (approved):

gularsson (changes requested):
Clicked wrong, and accidentaly approved. 
Great feature, works good!
However there are are some changes that i suggest. Mostly due to design regressions with the Avatar component.

src/components/AvatarGroup.tsx
dannelundqvist: Yes, that was the intention, but forgetness...

gularsson (approved):
Great! :heavy_check_mark:

PR #123, opened 2024-01-25
feat: use origin for opening new view

gularsson:


dannelundqvist (approved):
Now the navigation feels 💯 !

PR #128, opened 2024-01-29
refactored auth/refresh to fix issues where jwt is sometimes missing after login

dannelundqvist:


gularsson (approved):
Looks good, feels great!

PR #162, opened 2024-02-08
Feature/ele 392 investigate could awareness be used to signal which users

dannelundqvist:
Created backend part to track which user has which documents (planning, article, anything) open in a yjs tracker document. Client part is not finished as waiting for the useYMap to useYObserver change.


gularsson (approved):
Awesome!

PR #163, opened 2024-02-09
feat: parse planning

gularsson:


src-srv/utils/CollaborationServer.ts
```diff
@@ -205,7 +206,7 @@ export class CollaborationServer {
    */
   async #fetchDocument({ documentName: uuid, document: yDoc, context }: fetchPayload): Promise<Uint8Array | null> {
     // Init tracking document. Must not be fetched from cache when starting up the server fresh.
-    if (uuid === 'document-tracker' && !this.#openDocuments) {
+    if (uuid === 'document-tracker') {
```
dannelundqvist, line 209: I think we need to discuss this logic a bit more. This will reset the local #openDocuments every time a new user fetches the _document-tracker_ document. It might be that we need to handle this special document the same as other documents and allow it to be fetched from cache...

src-srv/utils/transformations/yjs/yPlanning.ts
```diff
@@ -0,0 +1,69 @@
...
+      }
+    })
+
+    groupedObjects[(object[groupKey]) as string].push(newObj as Block)
+  })
+
+  return groupedObjects
+}
+
+// TODO: Almost identical to newsDocToYmap function, make reuseable?
```
dannelundqvist, line 52: It seems at first glance as they are indeed almost the same. It would maybe be an idea to make this more generic and document it a little bit. Let's have a discussion around it.

src/lib/yMapValueByPath.ts
dannelundqvist, line 1: 🌮🌮🌮

dannelundqvist (changes requested):
Fantastic refactor/improvement!

I am a bit worried about how we handle the tracker-document in CollaborationServer (event without your change). And we should discuss it.

Putting that aside it's 💯

src-srv/utils/CollaborationServer.ts
gularsson: Yes, this needs to be discussed. Previous solution led to #fetchDocument trying to open `document-tracker` from repository, which threw errors. (unsure if that was just in dev server)

src-srv/utils/transformations/yjs/yPlanning.ts
gularsson: I'll look into that, tried to keep them separated at first so we wouldn't get any regressions in the handling of textbit editing. But I think we have some bugs (hopefully solved in this) in `newsDocToYmap`.

dannelundqvist (approved):

PR #196, opened 2024-02-26
Refactor yobserver

gularsson:


src/hooks/useYObserver.tsx
```diff
@@ -1,73 +1,109 @@
...
-export function useYObserver<T>(y: Y.Map<unknown> | undefined, key: string): [T, YSetter]
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export function useYObserver<T extends Record<string, any>>(y: Y.Map<unknown> | undefined, key?: string): [Record<string, T>, YSetter]
-export function useYObserver<T>(y: Y.Array<unknown> | undefined, key?: string): [T[], YSetter]
-export function useYObserver<T>(y: Y.Map<unknown> | Y.Array<unknown> | undefined, key?: string): [T | Record<string, T> | T[], YSetter] {
-  const [value, setValue] = useState<unknown | undefined>(get(y as Y.Map<unknown>, key))
+  // Get wanted Y.Map by path provided
+  const map = yMapValueByPath.get(yPlanning, path)
 
+  // Observe whole yPlanning to detect changes at top
```
dannelundqvist, line 36: Are the two comments misleading? It's not just planning we're tracking?

dannelundqvist (changes requested):
A thorough and solid refactor. I'll probably have more to say about it after using the code a bit more. Found two slightly misleading comments though. Other than that - great job!

src/hooks/useYObserver.tsx
gularsson: I would say it's accurate, it's just the planning/yPlanning we're observing. We're getting the planning ymap and attaching the observeDeep on that:
```
const yPlanning = document?.getMap('planning')
...
    yPlanning?.observeDeep((events) => {
```
Changes to `open-documents` and `content` shouldn't be observed afaik.

It's possible that the second observer (on the `map`) is superfluous. We should be able to catch all the changes on the observeDeep.

gularsson: Now i see, I'll fix so we also support the `article` Y.Map (and others)

dannelundqvist (approved):
Good solution, seems really easy to use now!

PR #241, opened 2024-03-12
fix: PlanningOverview margins

gularsson:


dannelundqvist (changes requested):
I would rather like to see padding/margin **inside** of the leftmost and rightmost cell. 
Skärmavbild 2024-03-12 kl  13 24 42

gularsson:
Nice drawing, yes that makes sense.

dannelundqvist (approved):
💯

PR #254, opened 2024-03-14
Feature/ele 434 upgrade textbit editor in elephant chrome

dannelundqvist:
Updated to latest textbit and textbit plugins.


gularsson (approved):
:1st_place_medal:

PR #259, opened 2024-03-15
Feature/ele 467 harmonize icons

gularsson:


dannelundqvist (approved):
Nice job. But there is a feeling it could be a good idea to have a a global css variable or something to controll style of all icons...

PR #261, opened 2024-03-15
Feature/ele 460 dropdown with check for document status

gularsson:


src/views/Planning/components/PlanDocumentStatus.tsx
dannelundqvist, line 1: Would it be possible to move styling into the combobox instead of specifying it in the usage of the component? Would make it easier to make changes to styling in the future.

src/views/Planning/components/PlanStatus.tsx
dannelundqvist, line 1: Same thing here. Font and other styling that should be the same in all usage could be handled inside the component for an easier future.

dannelundqvist (changes requested):

dannelundqvist (approved):

PR #266, opened 2024-03-18
Feature/ele 473 fix planslugline flicker buttoneditable make general

gularsson:


dannelundqvist (approved):
Things like text-muted-foreground should preferrably not be part of classNames, it should rather be decided by which "variant" is used. Not feasible as part of this PR to change it though.

PR #267, opened 2024-03-18
fix: click setActive and op or noop

gularsson:


gularsson:
Doesn't work :100: yet

dannelundqvist (changes requested):
I think this might be a lot simpler if you just use an onClick handler in the capture phase (with stopPropagation() for the badge, and then a normal (bubble) phase click event handler for the row. Then a click on the badge would simply be caught before the click on the row. The result would be that neither ref or special checks on the target would be needed in the handleSetActive func. What do you think?

dannelundqvist (approved):
As we discussed, we ignore this for later.

dannelundqvist (approved):

PR #288, opened 2024-03-25
fix: create new items

gularsson:


gularsson:
Implemented in (partially?) in https://github.com/ttab/elephant-chrome/commit/31e3242104eaeb6d6816f571aac2a59fe9349155
Might be some bug fixes that we need to keep.

src/lib/yMapValueByPath.ts
dannelundqvist, line 1: A wee bit picky maybe. But this file has a lot of nested if clauses which makes it a bit harder to follow the logic and, I'd argue, also makes it easier to introduce bugs.

Maybe we should try to return early more? Like
```
if (current instanceof Y.Map) {
  current.set(keys[keys.length - 1], value)
  return
}

if (current instanceof Y.Array === false) {
  return
}

if (isNaN(index)) {
  throw new Error(`Invalid array index '${keys[keys.length - 1]}'`)
}

if (current.length - 1 >= index) {
  current.doc?.transact(() => {
    current.delete(index)
    current.insert(index, [value])
  })
  return
}

// ... and more...
```

dannelundqvist (changes requested):

dannelundqvist (approved):

PR #299, opened 2024-03-28
fix(Planning): separate awareness events from CollaborationContextPro…

gularsson:
…vider


dannelundqvist (approved):
Seems to be a really nice solution to this issue!

PR #300, opened 2024-03-28
fix(Navigation): reset navigation

gularsson:
Reset by clicking title or icon in ViewHeader or by keyboard shortcut alt+shift+<view number> this combination is used due to its resemblance to alt+<number> which is browser default to switch tabs


dannelundqvist (approved):
Works like a charm! 💯

PR #326, opened 2024-04-10
Feature/ele 508 convert yplanning to newsdoc

gularsson:


dannelundqvist (approved):
There is room for typescript improvement. But... only if you feel you have the energy :)

PR #337, opened 2024-04-15
Feature/ele 517 handle yxmltext in newsdoc y conversion

gularsson:


src-srv/utils/transformations/yjs/yPlanning.ts
dannelundqvist, line 1: Should we not rename this file as it is more generic than just "planning"?

dannelundqvist (changes requested):
Fantastic job! A lot cleaner in many places. Only one question about a file name.

src-srv/utils/transformations/yjs/yPlanning.ts
gularsson: Yes, certainly, will do.

dannelundqvist (approved):

PR #341, opened 2024-04-16
Feature/ele 513 planning view header should adapt to being in a dialog

dannelundqvist:


gularsson (approved):
:+1: nothing to complain about :slightly_smiling_face:

PR #345, opened 2024-04-17
test: add tests for transformation slate to newsdoc

gularsson:


src-srv/utils/CollaborationServer.ts
```diff
@@ -218,13 +218,14 @@ export class CollaborationServer {
...
     // Fetch content
-    const { document = null } = await this.#repository.getDoc({
+    const { document = null, ...rest } = await this.#repository.getDoc({
       uuid,
       accessToken: context.token
     }) || {}
 
 
     if (document) {
+      console.log(rest)
```
dannelundqvist, line 228: Leftover that should be removed?

src-srv/utils/transformations/newsdoc/core/text.ts
```diff
@@ -26,9 +26,10 @@ const replace = (type: string, translateList: Record<string, string>): string =>
 }
 
 const createDocument = (): Document => {
+  // TODO: does this work, we'll find outj
```
dannelundqvist, line 29: It seems like it's working? Should the comment be removed?

dannelundqvist (approved):
Good tests! Two cleanliness questions.

src-srv/utils/transformations/newsdoc/core/text.ts
gularsson: Next attempt/PR is to revert and save. Then we'll find out

PR #362, opened 2024-04-24
feat: snapshot extension

gularsson:


dannelundqvist (changes requested):
Looks good and simple!

Lodash though, is more or less deprecated in many respects. Would it be possible to use `src/lib/debounce.ts` - and make it possible to cancel it -  instead of pulling in lodash as a new dependency?

gularsson:
> Looks good and simple!
> 
> Lodash though, is more or less deprecated in many respects. Would it be possible to use `src/lib/debounce.ts` - and make it possible to cancel it - instead of pulling in lodash as a new dependency?

I agree, I had just removed all lodash uses from the repo, though this application felt like a good compromise do reduce complexity, especially since we're doing it async...

dannelundqvist (approved):
As discussed. We'll leave the lodash dep in for now.

PR #384, opened 2024-05-06
feat: update connection with refreshed token, shared dir

gularsson:


src-srv/utils/extension-auth.ts
dannelundqvist, line 1: Perhaps it would be better to name the file "Auth.ts" or if that is unclear maybe "AuthExtension.ts"? It would follow the pattern of the other files.

dannelundqvist (changes requested):
Well needed cleanup. Nice! Stateless looks nice too. Just would like to suggest a filename change.

src-srv/utils/extension-auth.ts
dannelundqvist: An extension dir would probably be nice. We might create more extensions later.

PR #385, opened 2024-05-06
fix: transact version change, identify agent of change

gularsson:


src-srv/utils/extension-snapshot.ts
dannelundqvist, line 1: Is the console.log() a debug leftover? I would suggest we use console.info(), .warn(), etc for logging when we intentionally want it. Shows intent better than generic .log().

dannelundqvist (changes requested):

src-srv/utils/extension-snapshot.ts
gularsson: Misstake, removed.

dannelundqvist (approved):

PR #410, opened 2024-05-15
feat: verify no changes, filter __inProgress Ymaps

gularsson:


src-srv/utils/CollaborationServer.ts
```diff
@@ -231,6 +219,54 @@ export class CollaborationServer {
...
+    // Convert yDoc to newsDoc, so we can hash it and maybe save it to the repository
+    const { document } = yDocToNewsDoc(yDoc)
+
+    const currentHash = createHash(JSON.stringify(document))
+
+    // Compare original hash with current hash to establish if there are any changes
+    // This solution relies on the same order of keys in the documents, but a change of key order
+    // should be indicate a change in the document anyway.
+    if (yDoc.getMap('hash').get('hash') === currentHash) {
+      console.debug('::: saveDoc: No changes in document')
```
dannelundqvist, line 240: Is this intentionally left in the code? Will produce a lot of log messages later on.

src-srv/utils/CollaborationServer.ts
```diff
@@ -231,6 +219,54 @@ export class CollaborationServer {
...
+      })
+
+      await connection.transact(doc => {
+        const versionMap = doc.getMap('version')
+        const hashMap = doc.getMap('hash')
+        versionMap.set('version', result?.response.version.toString())
+        hashMap.set('hash', currentHash)
+      })
+
+      console.debug('::: Snapshot saved: ', result.response.version, 'new hash:', currentHash)
```
dannelundqvist, line 266: Is this intentionally left in the code? Will produce a lot of log messages later on.

dannelundqvist (changes requested):
I understand the reasoning behind debug logging. Maybe we should put those debug log statements inside an if checking (env !== prod)?

src-srv/utils/CollaborationServer.ts
gularsson: My reasoning is that we should have a logging strategy (and log library) and via envs be able to set our desired loglevel. debug, info, warn, error.
If in case we need to debug something in prod we can change to `logLevel: debug` and without code changes get a better understanding.

This _will_ create a lot of messages right now, but I'd like to see it as an indication on what type of log message this should be, when we have a strategy in place.

dannelundqvist (approved):

PR #424, opened 2024-05-17
feat: POC for saving after creation of planning

gularsson:


src/lib/planning/createPlanningDocument.ts
dannelundqvist, line 1: Should we not set "todays date" instead of hardcoded value from start?

src/lib/planning/search.ts
```diff
@@ -52,6 +52,8 @@ export const search = async (endpoint: URL, jwt: JWT, params?: SearchPlanningPar
     sort
   }
 
+  console.log(JSON.stringify(query, null, 2))
```
dannelundqvist, line 55: Intentional?

dannelundqvist (changes requested):

src/lib/planning/search.ts
gularsson: not at all

src/lib/planning/createPlanningDocument.ts
gularsson: yes

dannelundqvist (approved):

PR #449, opened 2024-05-27
Observe Y.XmlText values, don't allow empty title (or empty slug for …

dannelundqvist:
…text assignment)


src/hooks/useYValue.tsx
```diff
@@ -0,0 +1,127 @@
...
+ * @param deep boolean
+ *
+ * @returns [<T>, YParent]
+ */
+export function useYValue<T>(path: string, deep: boolean = false): [
+  T | undefined,
+  YParent
+] {
+  const { provider, synced } = useCollaboration()
+  const yRoot = synced ? provider?.document.getMap('ele') : undefined
```
gularsson, line 24: There was something about synced... It only allows server changes, not offline...?

gularsson (approved):
Great work, I feel that we could probably replace the useYObserver for this one eventually. :muscle:

dannelundqvist:
Also implemented setter so that the useYValue() hook now works like useState(). Regardless of whether the values parent container is a Y.Map or Y.Array.

Example usage:

```javascript
const [assignmentType, setAssignmentType] = useYValue<string | undefined>(`meta.core/assignment[${index}].meta.core/assignment-type[0].value`)
...
setAssignmentType('graphic')
```

gularsson (approved):
:+1:

PR #462, opened 2024-06-03
Feature/ele 588 planning item assignment list improvements

dannelundqvist:
Added useLink hook to simplify opening views.
Added generic prompt dialog to simplify using prompt modals.
Use useYValue() instead of useYObserver() to simplify

### DIFF
--- a/src/components/DataItem/AssignmentType.tsx
+++ b/src/components/DataItem/AssignmentType.tsx
@@ -17,7 +17,7 @@ export const AssignmentType = ({ path, editable = false }: {
     return <></>
   }
 
-  const value = state.map((s) => s.value).join('/')
+  const value = state.map((s) => s.value).sort().join('/')
 
   const selectedOption = AssignmentTypes.find(type => type.value === value)
 
--- a/src/components/DataItem/AssignmentType.tsx
+++ b/src/components/DataItem/AssignmentType.tsx
@@ -17,7 +17,7 @@ export const AssignmentType = ({ path, editable = false }: {
     return <></>
   }
 
-  const value = state.map((s) => s.value).sort().join('/')
+  const value = state?.map ? state.map((s) => s.value).sort().join('/') : ''
 
   const selectedOption = AssignmentTypes.find(type => type.value === value)
 
--- a/src/hooks/useLink.tsx
+++ b/src/hooks/useLink.tsx
@@ -1,8 +1,9 @@
 import type { KeyboardEvent, MouseEvent } from 'react'
 import { useNavigation, useView } from '@/hooks'
 import { handleLink } from '@/components/Link/lib/handleLink'
+import type { View } from '../types'
 
-export const useLink = (viewName: 'Editor' | 'PlanningOverview' | 'Planning') => {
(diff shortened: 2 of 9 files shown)


src/components/DataItem/AssignmentType.tsx
```diff
@@ -17,7 +17,7 @@ export const AssignmentType = ({ path, editable = false }: {
     return <></>
   }
 
-  const value = state.map((s) => s.value).join('/')
+  const value = state.map((s) => s.value).sort().join('/')
```
gularsson, line 20: :+1:

src/hooks/useLink.tsx
```diff
@@ -0,0 +1,25 @@
+import type { KeyboardEvent, MouseEvent } from 'react'
+import { useNavigation, useView } from '@/hooks'
+import { handleLink } from '@/components/Link/lib/handleLink'
+
+export const useLink = (viewName: 'Editor' | 'PlanningOverview' | 'Planning') => {
```
gularsson, line 5:
```suggestion
import { View } from 'what-ever-path'
export const useLink = (viewName: View) => {
```
Would this work? Then we don't need to change the types when/if we add views?

src/views/Planning/components/AssignmentRow.tsx
```diff
@@ -1,288 +1,224 @@
...
+      <AssignmentRowContent
+        index={index}
+        setSelectedAssignment={setSelectedAssignment}
+      />
+    </div>
   )
 }
 
-const AssignmentRowContent = ({ index, setSelectedAssignment, setShowCreateDialog, showCreateDialog = false }: {
+const AssignmentRowContent = ({ index, setSelectedAssignment /*, setShowCreateDialog, showCreateDialog = false */ }: {
```
gularsson, line 43: Remove comment?

src/views/Planning/components/AssignmentRow.tsx
```diff
@@ -1,288 +1,224 @@
...
-  setShowCreateDialog?: React.Dispatch<React.SetStateAction<boolean>>
-  showCreateDialog?: boolean
 }): JSX.Element => {
-  const { get } = useYObserver('meta', `core/assignment[${index}]`)
-  const { get: getAssignmentDescription } = useYObserver('meta', `core/assignment[${index}].meta.core/description[0].data`)
-  const { get: getAssignmentPublishTime } = useYObserver('meta', `core/assignment[${index}].data`)
+  const { provider } = useCollaboration()
+  const openArticle = useLink('Editor')
+  const base = `meta.core/assignment[${index}]`
+  // const [id] = useYValue<string>(`${assPath}.id`)
```
gularsson, line 50: Remove?

gularsson (changes requested):
useLink and Prompt looks great, some minor things.

src/hooks/useLink.tsx
dannelundqvist: Yes, that is a good idea! Otherwise this will be a headache in the future. I'll add this as a separate issue though,

gularsson (approved):
👍

PR #464, opened 2024-06-03
Added simple error view, loaded on trying to load non existant view

dannelundqvist:
Works only on simple cases for now. Should be generalized later on.


gularsson (approved):
Great. As you said, this would probably eventually be a Error boundary.

PR #465, opened 2024-06-03
Make ellipsis button standout better in hovered rows

dannelundqvist:


gularsson (approved):
This PR looks good, however I would argue the choice of naming it `bg-accent2`. There has to be a better name? If not, `bg-accent-2` is more tailwindish.

PR #485, opened 2024-06-07
feat: use vite for express

gularsson:


dannelundqvist (approved):
Good work finding a solution to this!

PR #494, opened 2024-06-11
Feature/ele 612 use keycloak

gularsson:


src-srv/utils/authenticatedUser.ts
dannelundqvist, line 1: File and function should be renamed to better sync with what it does.

src-srv/utils/extensions/auth.ts
```diff
@@ -1,40 +1,27 @@
...
-  constructor(configuration: Configuration) {
-    this.configuration = { ...configuration }
-  }
-
   async onAuthenticate({ token }: onAuthenticatePayload): Promise<{
     token: string
     user: JWTPayload
   }> {
-    try {
+    /* try {
```
dannelundqvist, line 12: Is this a leftover commented out code or WiP that should be there?

src-srv/utils/extensions/auth.ts
```diff
@@ -1,40 +1,27 @@
...
+    } */
 
     return {
       token,
       user: { ...decodeJwt(token) }
     }
   }
 
-  async onStateless({ payload, connection }: onStatelessPayload): Promise<void> {
+  /* async onStateless({ payload, connection }: onStatelessPayload): Promise<void> {
```
dannelundqvist, line 24: Is this a leftover commented out code or WiP that should be there?

src/views/PlanningOverview/PlanningGrid.tsx
```diff
@@ -35,7 +36,7 @@ export const PlanningGrid = ({ startDate, endDate }: PlanningGridProps): JSX.Ele
   }, [startTime, endTime, indexUrl, locale])
 
   const { data } = useSWR(searchUrl.href, async (): Promise<PlanningsByDate | undefined> => {
-    if (!jwt) {
+    if (status !== 'authenticated' || !session?.accessToken) {
```
dannelundqvist, line 39: We will do this a lot. Even though it is little code this could be a strong candidate for a function call like `!isValidSession(session) { return }`. Would make it readable, easier to extend/change and...

dannelundqvist (changes requested):
Fantastic job! Looks like this has cleaned up the code and logic quite a bit. Only have a few questions/minor things.

src-srv/utils/extensions/auth.ts
gularsson: Yeah, it should be reimplemented... I need to take a look at it right afterwards

src-srv/utils/extensions/auth.ts
gularsson: Yeah, it should be reimplemented... I need to take a look at it right afterwards

src/views/PlanningOverview/PlanningGrid.tsx
gularsson: > We will do this a lot. Even though it is little code this could be a strong candidate for a function call like `!isValidSession(session) { return }`. Would make it readable, easier to extend/change and...

Yes, the reason here was mostly because of types, I think `session` is `Session | undefined` but if we are `authenticated` we should also have an `accessToken`, so it feels a bit superfluous. I need to look more closer to this.

dannelundqvist (approved):

PR #496, opened 2024-06-11
Feature/ele 596 empty textboxes are not rendered no placeholder shown

dannelundqvist:
Major improvements to useYValue() which enables us to (almost) automatically create missing structures like tt/slugline which is often missing. I think we now have all the pieces to replace useYObserver() completely.


src/hooks/useYValue.tsx
```diff
@@ -31,13 +43,28 @@ export function useYValue<T>(path: string): [
...
+      setValue(isYXmlText(vp[0]) ? vp[0].toJSON() as T : vp[0] as T)
+    } else {
+      // Just return raw Y values when not observing
+      setValue(vp[0] as T)
+    }
 
-    if (isYContainer(p)) {
-      setParent(p)
+    if (isYContainer(vp[1])) {
+      setParent(vp[1])
```
gularsson, line 65: I'm not too found of this, I'd rather see something along the line as the previous (and even more verbose):
`const [value, path] = getValueByPath(...)`

No one char variables, and no magic numbers.

Though I see now that you're reassigning it, which kind makes it harder.
I'm not too found of reassigning either :wink:

gularsson (approved):
One thing to consider

src/hooks/useYValue.tsx
dannelundqvist: Good points. I Agree. I'll refactor it!

Victor-Lindh (approved):

gularsson (approved):
:+1:

PR #501, opened 2024-06-13
Add kc_idp_hint to authorization URL

fred-o:
A configurable hint value makes it possible to skip the Keycloak login dialog and go directly to the Identity Provider


gularsson (approved):
:+1:

PR #502, opened 2024-06-13
feat: use tt/visual

gularsson:
This should do it?

### DIFF
--- a/src-srv/utils/transformations/newsdoc/tt/visual.ts
+++ b/src-srv/utils/transformations/newsdoc/tt/visual.ts
@@ -7,16 +7,17 @@ export const transformVisual = (element: Block): TBElement => {
   return {
     id,
     class: 'block',
-    type: 'core/image',
+    type: 'tt/visual',
     properties: {
-      // FIXME: This is a hack to get a viewable image
-      // TODO: Could there be more than one link/image?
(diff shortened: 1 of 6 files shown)


src-srv/utils/transformations/newsdoc/tt/visual.ts
```diff
@@ -7,34 +7,38 @@ export const transformVisual = (element: Block): TBElement => {
...
+      // FIXME: This is a hack to get a viewable image,
+      // we need authentication to view non-watermarked images
+      href: links[0]?.url
+        ?.replace('_NormalPreview.jpg', '_WatermarkPreview.jpg') ?? '',
       rel: links[0].rel,
       uri: links[0].uri,
       type: links[0].type,
       credit: links[0].data.credit,
-      altText: data.caption,
+      altText: data?.altText,
```
dannelundqvist, line 20: Should it be set to undefined rather than empty string?

dannelundqvist (approved):

src-srv/utils/transformations/newsdoc/tt/visual.ts
gularsson: After consideration, all values should be defined as their unset representation in Go. Ie string '', number 0 etc.
altText isn't a valid newdoc attribute so i've removed it. We should decide if we should have it.

PR #520, opened 2024-06-17
Feature/ele 623 get rid of sign in with keycloak page

gularsson:


src/components/App/Menu.tsx
```diff
@@ -51,7 +51,12 @@ export const Menu = (): JSX.Element => {
             </SheetClose>
 
             <SheetClose asChild>
-              <a href={`${import.meta.env.BASE_URL}/api/auth/signout`}>
+              <a href='' onClick={(event) => {
+                event.preventDefault()
+                signOut()
+                  .catch((error) => console.log(error))
```
dannelundqvist, line 57: Shouldn't we do something when this happens? I understand it is not that important, but still?

dannelundqvist (approved):
Nice work!

Thinking about our recent discussions around catching and handling errors when seeing the console.log() in the catch with no further handling. Does errors like that need to be handled/communicated somehow? If you deem it is is pointless to do something else, then just leave it as is.

src/components/App/Menu.tsx
gularsson: Yes, we really should do something. But I think that's a whole different task. We should show this to the user. I guess console.error() would be better but in the end it's the same result.

I'll change to console.error so we'll find it once we'll get to fixing it.

PR #524, opened 2024-06-18
build: use vitest as test runner

gularsson:


dannelundqvist (approved):
Looks good. I  don't have any objections :)

PR #525, opened 2024-06-18
fix: types, urls, style etc

gularsson:


src/plugins/ImageSearch/index.tsx
```diff
@@ -16,14 +17,14 @@ function handler({ dispatch, viewRegistry, origin, id }: ActionHandlerI): boolea
...
   const { viewId: origin } = useView()
   const viewRegistry = state.viewRegistry
   const query = useQuery()
   const { id } = query
 
-  return ({
+  return {
     class: 'block',
     name: 'tt/visual/search',
     componentEntry: {
```
gularsson, line 30: Unsure what the proper type is here:
It says we're missing `component` in `componentEntry`. Not sure if we're missing something.

Victor-Lindh: the `component` is the acting parent/wrapper, if I understand it correctly. In this case, there isn't a parent component.

dannelundqvist: You can either define the function as a Plugin.IniFunction. Or define it to return a Plugin.Definition. Either way is fine. But it does not matter for the issue at hand.

A componentEntry must always have a component. But this plugin does not need a componentEntry. So remove that entirely.

PR #600, opened 2024-08-08
Feature/ele 584 planning item story picker should use real data

dannelundqvist:
Refactored all code that handles IndexedDB to better work with multiple document types. Force refresh of data at least every 48 hours. IndexedDBProvider creates all object stores upon first load or when version is increased.


src/views/Planning/components/PlanStory.tsx
Victor-Lindh, line 1: While it's possible to add and change a set story, it doesn't seem to be a way to remove one from a planning item?

dannelundqvist: I know. It is part of a specific combobox improvement issue (#ele-672).

Victor-Lindh (approved):

PR #601, opened 2024-08-08
Feature/ele 578 planning item publisheddraft picker should work correctly

Victor-Lindh:


src/App.tsx
dannelundqvist, line 1: Does the PlanningTableProvider really need to be available in the whole application? I might miss something but it feels like something like this provider should be as close to the usage as possible (and not be nested).

dannelundqvist (changes requested):
I would like to discuss the structure around planning table provider. Other than that it looks good!

src/App.tsx
Victor-Lindh: Yes, correct. It's was unnecessary, and didn't add anything.
I've removed it in a new commit.

dannelundqvist (approved):

PR #603, opened 2024-08-09
Feature/ele 582 section sector picker should use real data

dannelundqvist:
Replaced all uses of (hardcoded) section default values with cached data fetched from the index.

This meant we lost "color dots" for sections in tables. That data does not exist in the index. We should discuss that as a separate issue. (I left the defaults files as deprecated reference for this reason)


src/lib/index/authors.ts
gularsson, line 1: `authors.ts`, `sections.ts` and `stories.ts` are close to identical. One function? Could maybe just use `get.ts`?

src/lib/index/stories.ts
gularsson, line 1: See `authors.ts`

src/datastore/contexts/CoreSectionProvider.tsx
gularsson, line 1: Refactor? See PR comment at top

src/hooks/useSections.tsx
gularsson, line 1: Refactor? See PR comment above.

src/views/Planning/components/PlanSection.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+export const PlanSection = (): JSX.Element => {
+  const allStories = useSections().map((_) => {
+    return {
+      value: _.id,
+      label: _.title
+    }
+  })
+
+  const [, setStories] = useYValue<Y.Array<unknown>>('links.core/section', {
```
gularsson, line 16: `setStories`? Should be `setSections` right?

src/views/Planning/components/PlanSection.tsx
```diff
@@ -0,0 +1,53 @@
+import { Awareness } from '@/components'
+import { ComboBox } from '@/components/ui'
+import { useSections, useYValue } from '@/hooks'
+import { Block } from '@/protos/service'
+import { useRef } from 'react'
+import * as Y from 'yjs'
+
+export const PlanSection = (): JSX.Element => {
+  const allStories = useSections().map((_) => {
```
gularsson, line 9: `allSections`

gularsson (changes requested):
Over all looks good, great to finally get proper data to work with and get rid of the mocked/static data

I cant help to wonder if we shouldn't start to refactor this immediately. It's easy for me to say coming from an outside perspective :slightly_smiling_face: 

`CoreSectionProvider`, `CoreAuthorProvider` and `CoreStoryProvider` are (close to?) identical.

`useAuthor`, `useSection` and `useStory` could be one hook.
```
const { authors, section, story } = useIndexedData()
```
I dont see any point in keeping them as separate contexts or hooks in terms of rerenders or codewise.

Victor-Lindh:
Besides the points that @gularsson already brought up above, I have nothing further to add.
Looks great otherwise!

dannelundqvist:
> Over all looks good, great to finally get proper data to work with and get rid of the mocked/static data
> 
> I cant help to wonder if we shouldn't start to refactor this immediately. It's easy for me to say coming from an outside perspective 🙂

Totally agree, over the course of doing refactoring the originally messy code clearer patterns have emerged. On it. It will saves us tons of time later on.

src/datastore/lib/fetchSections.ts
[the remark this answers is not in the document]
gularsson: I saw that we lost the section colors, a specific section "transformer" could augment the data received from the index.

src/datastore/contexts/CoreSectionProvider.tsx
dannelundqvist: I will leave this as is for now.

src/hooks/useSections.tsx
dannelundqvist: Will leave as useSections is not the same logic as useAuthors for example.

PR #605, opened 2024-08-12
Fix useyvalue

gularsson:
This should be a step forward in being able to use useYValue in more contexts, and eventually replace useObserver. There might still be cases that are not supported, but I think we're better suited to solve those with a rigid test suit.


dannelundqvist (approved):
Looks good. Looking forward to be using this improvement elsewhere!

PR #606, opened 2024-08-12
refactor(tables): reduce, reuse

gularsson:


dannelundqvist (approved):
Very nice work in generalizing the table handling!

PR #612, opened 2024-08-14
Swedish filter

hazone:


hazone:
Fix for 
https://linear.app/tt-nyhetsbyran/issue/ELE-686/filtering-by-planning-status
https://linear.app/tt-nyhetsbyran/issue/ELE-684/filter-options-should-be-in-swedish

dannelundqvist (approved):

PR #639, opened 2024-08-21
feat: get organisations from index and cache

gularsson:


dannelundqvist (changes requested):
Unfortunately it crashes on first installation as the new object store, `core/organisation`, is not defined in `src/defaults/indexedDBSpecification.ts` it all looks great!

dannelundqvist (approved):

PR #646, opened 2024-08-23
feat: create structure and basic event view

gularsson:
TODO as other issues:
Datetime handling, currently in progress
~Registration, questions raised about its placement i newsdoc~ probably keep it like it is?
Connect planning from event
Show expanded organiser information? How?
General looks


src/views/Event/index.tsx
```diff
@@ -0,0 +1,171 @@
...
+            </div>
+
+            <Description role='public' />
+            <Registration />
+
+            <p>Datetime TODO</p>
+          </div>
+
+          <div className='flex flex-col space-y-2 w-fit'>
+            <div className='grid grid-cols-2 gap-2 @xs:flex'>
```
dannelundqvist, line 130:
```suggestion
            <div className='flex flex-wrap gap-2'>
```

src/views/Event/index.tsx
dannelundqvist, line 1: It is not possible to clear category.

dannelundqvist (changes requested):
Really good work. Nice that we generalize many components to be more reusable now.

Two small issues in the actual event dialog though. Category combobox issue and a suggested improvement to make the comboboxes work better when screen estate get small.

src/views/Event/index.tsx
gularsson: Does that work for you? Cause that was my first try, but it didn't work as expected :thinking: 
I'll try again.

src/views/Event/index.tsx
gularsson: :+1:

dannelundqvist (approved):

PR #666, opened 2024-09-05
Feature/ele 740 upgrade textbit and textbit plugins in elephant chrome

dannelundqvist:
Upgraded textbit-plugins so that we can remove text style role (previously type) translations etc.


gularsson (approved):
Nice :+1:

PR #671, opened 2024-09-06
Slugline tab order, improved documentStatus handling

dannelundqvist:
Tab order for slugline in planning was not working when switching between button and text input. Removed the button and restyled the context of the input to mimick previous behaviour. We should not allow changing slugline on usable planning.

Also added useDocumentStatus() hook that allows both getting and setting the document status much easier.

### DIFF
--- a/src/views/Planning/components/PlanDocumentStatus.tsx
+++ b/src/views/Planning/components/PlanDocumentStatus.tsx
@@ -1,54 +1,19 @@
 import { Awareness } from '@/components'
 import { ComboBox } from '@/components/ui'
 import { DocumentStatuses } from '@/defaults'
-import { useSession } from 'next-auth/react'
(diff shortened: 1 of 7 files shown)


src/views/Planning/components/PlanDocumentStatus.tsx
```diff
@@ -57,7 +22,12 @@ export const PlanDocumentStatus = ({ documentId }: { documentId: string }): JSX.
         options={DocumentStatuses}
         variant={'ghost'}
         selectedOption={selectedOption}
-        onSelect={(option) => handleOnSelect(option, documentStatus)}
+        onSelect={(option) => {
+          console.log('setting to ', option.value)
```
gularsson, line 26: :broom:

gularsson (approved):

PR #683, opened 2024-09-12
fix: update to use @ttab/elephant-api

gularsson:
Use generated typescript definitions from protos from a npm package... ~waiting for package build~


src-srv/utils/CollaborationServer.ts
```diff
@@ -305,10 +305,12 @@ export class CollaborationServer {
     }
 
     // Fetch content
+    console.log('::: uuid', uuid)
```
dannelundqvist, line 308: Maybe use `console.debug()` instead.

src-srv/utils/CollaborationServer.ts
```diff
@@ -305,10 +305,12 @@ export class CollaborationServer {
...
     const newsDoc = await this.#repository.getDoc({
       uuid,
       accessToken: context.accessToken
     }).catch(ex => {
+      console.log('Error fetching document', ex)
```
dannelundqvist, line 313: Maybe use console.error() instead.

gularsson: Left behind, will remove

src-srv/utils/CollaborationServer.ts
gularsson: Left behind, will remove

dannelundqvist (approved):

PR #686, opened 2024-09-13
Feature/ele 591 when creating an assignment the user should be helped in

dannelundqvist:
This branch blew up...
* Revamped how useYValue() works. It no longer automatically injects missing block structures. It now uses `useSyncExternalStore)` with deep comparison of structures to be more performant (avoid unnecessary rerenders). It does no longer return parent. Instead use useYValue('path', true) to explicitly get a parent that you can manipulate. (See _handles __inProgress_ test in `__tests__/useYValue.test.tsx` for example.)
* Improved yUtils: revamped `getValueByYPath()` to not auto create missing blocks, added `setValueByYPath()` and `deleteByYPath()`. They can all now take either string paths or Y array paths.
* Fixed many issues in how AssignmentTable and assignment handling works. Improved UI and UX. Now you can either edit an existing assignment or create a new, not both at the same time.
* Automatically set text assignment slug to planning slug if the slug has not been used in another text assignment.


gularsson (approved):
Comprehensive :+1:

Victor-Lindh (approved):

PR #687, opened 2024-09-13
Feature: Factbox views

Victor-Lindh:


src/lib/index/index.ts
dannelundqvist, line 1: The ts-expect-error hides a real issue which should not be fixed. We are exporting "search" from both ./plannings and ./factbox leading to ambiguity. Looking at the code it seems none of the "search" symbols should be directly exported as they are both available through the exported Planning and Factbox objects. We should revise this so it is less confusing.

src/views/FactboxEditor/index.tsx
dannelundqvist, line 1: We should create a Linear issue, **for later**, to refactor both the article and factbox editor as they have lots of child components that are the same.

src/views/FactboxSearch/formatModified.ts
dannelundqvist, line 1: I think this could be a general library/util/helper function that sits next to or with the other time/date util functions we have. Could be useful in many places.

dannelundqvist (changes requested):
Nice work!

I only have two improvement wishes.

src/views/FactboxEditor/index.tsx
```diff
@@ -0,0 +1,295 @@
...
+import { type YXmlText } from 'node_modules/yjs/dist/src/internals'
+import { TextBox } from '@/components/ui'
+import { Button } from '@ttab/elephant-ui'
+import { createStateless, StatelessType } from '@/shared/stateless'
+import { useSession } from 'next-auth/react'
+import { createDocument } from '@/lib/createYItem'
+import { factboxDocumentTemplate } from '@/lib/templates/factboxDocumentTemplate'
+
+const meta: ViewMetadata = {
+  name: 'FactboxEditor',
```
gularsson, line 23: Raising the question whether this is the path we want to use, it doesn't really adhere to the rest plannings planning events event etc (unsure of what we call the Image-paths). Is factbox enough? It might be a better path.

factboxes and factbox?

src/views/FactboxEditor/index.tsx
Victor-Lindh: Linear issue created: https://linear.app/tt-nyhetsbyran/issue/ELE-768/refactor-article-and-factbox-editors-into-common-parent

src/lib/index/index.ts
Victor-Lindh: New commit pushed

src/views/FactboxSearch/formatModified.ts
Victor-Lindh: New commit pushed

src/views/FactboxEditor/index.tsx
Victor-Lindh: I'm for changing into a simpler format, as you suggest here.
Since they reside in the `templates`folder, simply calling them by their initial names (planning, event, factbox) should be enough.
I can add that to this PR if that's ok with you and @dannelundqvist ?

src/lib/templates/factboxDocumentTemplate.ts
```diff
@@ -0,0 +1,24 @@
...
+* @returns Document
+*/
+
+export function factboxDocumentTemplate(id: string): Document {
+  return Document.create({
+    uuid: id,
+    type: 'core/factbox',
+    uri: `core://factbox/${id}`,
+    language: 'sv-se',
+    title: 'Rubrik',
```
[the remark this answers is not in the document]
dannelundqvist: Isn't this solved in your backend validation changes @gularsson? The crash that is. Later we should validate and notify user in the UI as well of course.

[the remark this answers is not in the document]
gularsson: Yes, that should be handled, lets put it this way, more gracefully ;)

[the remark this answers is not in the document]
dannelundqvist: So this issue could be safely ignored for now?

dannelundqvist (approved):

src/views/FactboxEditor/index.tsx
Victor-Lindh: Now I understand that it's the name of the view itself you were talking about.
So:
FactboxEditor -> Factbox
FactboxSearch -> Factboxes
Sounds good to me, easy enough to change.
Made yet another quick PR if this is accepted:
https://github.com/ttab/elephant-chrome/pull/689

src/lib/templates/factboxDocumentTemplate.ts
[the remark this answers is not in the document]
Victor-Lindh: New commit pushed

src/views/FactboxEditor/index.tsx
gularsson: :heavy_check_mark: 
Lets skip the renaming of the templates (for now)

gularsson (approved):

PR #747, opened 2024-09-30
Feature/ele 801 show new items immediately in open views optimistic update

gularsson:
Got a bit out of hand... Will start with this:
* Memoized table
* Grouping headers, newsvalue
* useFetcher and SWRProvider
* Planning and Events are fetched similarly
* Fix memoized filters
* useDeferredValue - less glitches when loading
* Fix selected row handling and navigation
* Fetch all pages/items
* Optimistic update in a separate grouping header "New items"
* Yjs document to store and track users "New items" and other data

### DIFF
--- a/src-srv/index.ts
+++ b/src-srv/index.ts
@@ -138,6 +138,7 @@ export async function runServer(): Promise<string> {
     case 'production': {
       // Catch all other requests and serve bundled app
       app.use(pinoHttp({ logger }))
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
       app.use(BASE_URL || '', express.static(distDir))
       app.get('*', (_, res) => {
         res.sendFile(path.join(distDir, 'index.html'))


dannelundqvist:
Seems to be a fantastic change in stability and performance. 🥳

Generally also really nice with the table group headers. Great difference! I have a few suggestions. We don't use shadows to lift information anywhere else in the system unless to elevate popovers, modals etc, I think it should be removed.

Headers in tables are traditionally (iOS) smaller in height and less prominent than the table data rows. I suggest we make it a tiny bit smaller to differentiate it more.

See attached screenshot with the upper header having the suggested changes compared to the lower.

Skärmavbild 2024-10-08 kl  08 16 42

src/components/Table/index.tsx
```diff
@@ -81,77 +134,95 @@ export const Table = <TData, TValue>({
...
-            }
+    return deferredRows.map((row) => {
+      return (
+        <React.Fragment key={row.id}>
+          <TableRow className='sticky top-0 bg-muted'>
+            <TableCell colSpan={columns.length} className='pl-6'>
+              <div className='flex justify-between items-center flex-wrap'>
+                <div className='flex items-center space-x-2'>
+                  <span className='font-thin text-muted-foreground'>Nyhetsvärde</span>
+                  <span className='inline-flex items-center justify-center w-6 h-6 bg-background rounded-full ring-1 ring-gray-300 shadow-md'>
```
dannelundqvist, line 167: Can use `size-6` instead of `h-6 w-6`. Also suggesting to make it smaller and removing shadow (see screenshot in PR comment).

src/components/Table/index.tsx
```diff
@@ -81,77 +134,95 @@ export const Table = <TData, TValue>({
...
+                    {row.groupingValue as string}
+                  </span>
+                </div>
+                <div className='flex items-center space-x-2 px-6'>
+                  <span className='font-thin text-muted-foreground'>Antal</span>
+                  <span className='inline-flex items-center justify-center w-6 h-6 bg-background rounded-full ring-1 ring-gray-300 shadow-md'>
```
dannelundqvist, line 173: Can use `size-6` instead of `h-6 w-6`. Also suggesting to make it smaller and removing shadow (see screenshot in PR comment).

src/lib/events/search.ts
```diff
@@ -36,8 +34,8 @@ export const search = async (endpoint: URL, accessToken: string, params?: Search
           {
             range: {
               'document.meta.core_event.data.start': {
-                gte: start.toISOString(),
-                lte: end.toISOString()
+                gte: params?.where?.start,
```
dannelundqvist, line 37: If we actually call this method without providing start or end options for the where clause, which is ok according to the interface, this search should break, won't it?.

Start and end should either be required or the range part should be added only when they are supplied.

src/lib/index/searchIndex.ts
dannelundqvist, line 1: Why have the properties page, pages and pageSize been removed from the result interface? It is still added in the actual return value. (Which is really helpful when handling paginated results.)

src/lib/index/plannings.ts
```diff
@@ -36,8 +34,8 @@ const search = async (endpoint: URL, accessToken: string, params?: SearchPlannin
           {
             range: {
               'document.meta.core_planning_item.data.start_date': {
-                gte: start.toISOString(),
-                lte: end.toISOString()
+                gte: params?.where?.start,
```
dannelundqvist, line 37: If we actually call this method without providing start or end options for the where clause, which is ok according to the interface, this search should break, won't it?.

Start and end should either be required or the range part should be added only when they are supplied.

src/views/EventsOverview/EventsList.tsx
```diff
@@ -1,94 +1,40 @@
...
-  }, [startTime, endTime, indexUrl])
-
+  const { error } = useSWR(['Events', from, to, { withPlannings: true }])
+  const columns = useMemo(() => eventTableColumns({ sections }), [sections])
 
-  const { data } = useSWR(['eventitems', status, searchUrl.href], async (): Promise<SearchIndexResponse<Event> & { planningItem?: string } | undefined> => {
-    if (status !== 'authenticated') {
-      return
+  const onRowSelected = useCallback((row?: Event) => {
+    if (row) {
```
dannelundqvist, line 20: Do we really need to log selects/deselects?

dannelundqvist (changes requested):

src/lib/index/searchIndex.ts
gularsson: Previous fetch only fetched the first page, leaving us with an incomplete list. Now I'm aggregating all pages to one result. I figured that information about pages and page size would be superfluous then...

dannelundqvist (approved):

PR #761, opened 2024-10-10
Feature/ele 777 create article as flash or 2pa2

dannelundqvist:
Includes bugs and ugly code. Definitely a WiP!


gularsson (approved):
Not that bad :slightly_smiling_face:

PR #776, opened 2024-10-14
feat: propagate data from planning and assignment to created article

gularsson:


src/defaults/templates/articleDocumentTemplate.ts
```diff
@@ -1,17 +1,61 @@
 import { Block, Document } from '@ttab/elephant-api/newsdoc'
+import { ungroup } from '../../../src-srv/utils/transformations/groupedNewsDoc'
```
dannelundqvist, line 2: We really should refactor all these functions to shared, but maybe not now...

dannelundqvist (approved):

src/defaults/templates/articleDocumentTemplate.ts
gularsson: Yeah really

PR #777, opened 2024-10-14
feat: add validation Assignment creation form

gularsson:
- Added `path` and `className` props to `Title` component.
- Integrated `handleValidation` function in `Assignment` component.
- Replaced `TextBox` with `Title` in `Assignment` component.
- Added validation logic to ensure required fields are filled.


dannelundqvist (approved):
I like the simplicity of using the validation!

PR #805, opened 2024-10-17
feat: refactor auth and document fetching logic

gularsson:
Refactored the authentication and document fetching logic to use session from res.locals instead of fetching the token in each handler. Removed unused files and updated the routes accordingly. Added authSession middleware to handle session retrieval.


dannelundqvist (approved):
:crossed_fingers::skin-tone-2:

PR #839, opened 2024-10-30
fix: exclude auth paths when serving locally

gularsson:


dannelundqvist (approved):
Seems to work just fine!

Victor-Lindh (approved):

PR #844, opened 2024-10-31
fix: get correct view as linkTarget

gularsson:


Victor-Lindh (changes requested):
App crashes when switching to `Assignments` overview

gularsson:
> App crashes when switching to `Assignments` overview Skärmavbild 2024-11-04 kl 08 39 37

This isn't optimal, we should be able to derive valid views from type variable and not have to keep a lookup table. Other options that comes to mind, was more hacky.

Victor-Lindh (approved):

PR #847, opened 2024-11-01
Feature/ele 860 article editor should call spellchecking backend and hook

dannelundqvist:


src/lib/repository/checkSpelling.ts
gularsson, line 1: You should update `@ttab/elephant-api` to latest and use the types.

I also think that there's a client to use instead of fetch: `import { CheckClient } from '@ttab/elephant-api/spell'`. With the client we get a typed response, with good errors, and other niceties.

I think we've been missing that elsewhere previously as well.

gularsson: Might even make this file superfluous?

gularsson (changes requested):
Great work!

Besides form side-discussion:
* FF not setting initial focus correctly, at top instead of where clicked
* unable to chose spelling suggestion and replace misspelled word

There's only one improvement that i found.

src/lib/repository/checkSpelling.ts
dannelundqvist: FF not setting initial focus correctly is address in PR https://github.com/ttab/textbit/pull/209

gularsson (approved):

PR #853, opened 2024-11-06
Feature/ele 862 factbox editor should call spellchecking backend and hook

dannelundqvist:
Spellcheck in flash and textbox.
Fixed z-index on context menu in dialogs.
Display description on phrase suggestions.


shared/Spellchecker.ts
gularsson, line 1: Kinda feeling that this should be a hook, which in turn could get the locale and session, and we don't need to pass them as arguments. Or is this used in the backend as well?

gularsson (approved):

PR #863, opened 2024-11-11
feature: search view

Victor-Lindh:


gularsson (approved):
Looks great, two minor things, and the linting.
What we discussed on the meeting can, as we said, be improvement issues.

src/defaults/applicationMenuItems.ts
```diff
@@ -19,6 +19,11 @@ export interface ApplicationMenuItem {
 }
 
 export const applicationMenuItems: ApplicationMenuItem[] = [
+  {
+    name: 'Search',
+    label: 'Sök',
+    icon: SearchIcon
+  },
   {
```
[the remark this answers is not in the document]
Victor-Lindh: fixed in new commit

src/components/Table/index.tsx
```diff
@@ -142,15 +142,20 @@ export const Table = <TData, TValue>({
...
-            {deferredLoading
-              ? 'Laddar...'
-              : 'Inga planeringar funna.'}
+            {isSearchTable
+              ? deferredLoading
+                ? ''
+                : 'Inga poster funna.'
+              : deferredLoading
+                ? 'Laddar...'
+                : 'Inga planeringar funna.'}
```
gularsson, line 158: Since 'Inga planeringar funna' is the wrong phrase in most of occasions, we might as well change to 'Inga poster funna' for all? Its more general.

src/views/SearchOverview/PoolDropdown.tsx
```diff
@@ -23,7 +25,8 @@ export const PoolDropdown = ({ pool, setPool }: DropdownProps) => {
       options={pools}
       onSelect={(e) => {
         setPool(e.value)
+        setQueryString({ ...query, s: e.value })
```
gularsson, line 28: the setter is appending, so spreading query is unnecessary.
The key for the parameter should be discussed, I like explicit naming which means I've used from=, id= source= etc. It gives longer but clearer urls.

Either way, it's best if we're consistant

Victor-Lindh: should be fixed in new commit

src/components/Table/index.tsx
Victor-Lindh: fixed in new commit

gularsson (approved):

PR #868, opened 2024-11-12
feat: add useQuery hook with tests

gularsson:
- Implemented useQuery hook to manage URL query parameters.
- Added unit tests for useQuery hook to ensure functionality.


Victor-Lindh:
Should `setQueryString` perhaps encode the parameters?
Or is that up to the method calling the method?
Skärmavbild 2024-11-12 kl  14 40 57

gularsson:
> Should `setQueryString` perhaps encode the parameters? Or should that be done prior to calling the method? Skärmavbild 2024-11-12 kl 14 40 57

Seems like we're getting fooled by chrome, it's encoded already ;)

Victor-Lindh (approved):

PR #884, opened 2024-11-18
Ele 897 hidden auto assign user to flash

Victor-Lindh:


Victor-Lindh:
> It is good to guard against creating flashes when no user exist. But it would be frustrating to find out first when we have filled everything in.
> 
> We need to add an error message. Like the message in the attached image but visually an error box. The button should also be grayed out/disabled.

I addressed this in a complementary PR: https://github.com/ttab/elephant-chrome/pull/887
@dannelundqvist

gularsson (approved):
:+1: 
Pointed a PR to this branch for Form usage.

dannelundqvist (approved):

gularsson (approved):

PR #886, opened 2024-11-19
Feature/ele 901 keyboard navigation between views

dannelundqvist:
Revamped navigation code to:
* make it more stable as window.history is async
* allow arrow navigation between views <-, ->
* add ESC key shortcut to unfocus text fields to allow arrow navigation
* close views in the middle of multiple open views then rollback and rollup history navigation again but without the closed view
* open many views of the same type (e.g planning or editor or...) without adding them to the history which gives a more natural browser back button experience for the user
* Moved component rendering and view size calculationk from navigation initialization, reducer/provider to AppContent where views are actually rendered
* Way less code to manage navigation state


gularsson (approved):
Great work, nice cleanup :broom: 

Works great.

One issue, when navigating to an initial state using an url, two history entries seems to be written. I have to back twice to get back. Same with any url being written.

Victor-Lindh (commented):
Looks nice! Great job.
One issue, not sure if this is the same problem that Gustav had in an earlier comment:
1. From the plannings overview, click open a new planning.
2. Copy the url with the id in the querystring.
3. Open a new tab.
4. Paste and navigate to the url. For me, I'm being redirected back to /plannings.

dannelundqvist:
Loosing the query string was bad. It is fixed. The double history entry is in Firefox only and does not interfere much (if any) with the user experience. I'm ignoring that for now as it would require more work.

PR #887, opened 2024-11-19
flash: added error message, disabled send button if no author documen…

Victor-Lindh:
…t is found


gularsson (approved):
Might want to break out the Alert to a separate component, the nested ternary impairs the readability.

dannelundqvist (approved):

PR #892, opened 2024-11-19
Icon placement fix

dannelundqvist:


gularsson (changes requested):
I think we need pl-6 when asDialog true, the Dialog header got a bit askew.

gularsson (approved):

PR #901, opened 2024-11-21
Feature/ele 910 active view should be more visually prominent

dannelundqvist:
* new components View.Root and View.Content which _should be used for all views in the future_ as it wraps and hides tabs and scroll functionality to make it easier to have all views look and behave the same
* use new elephant ui that hides default close cross in dialogs by default so not to conflict with views in dialogs
* Improved "elevation" of active view
* gray header of active view
* elevate focused view
* improved "loading" texts as it's own reusable component for reusable design
* improved logout button design


gularsson (approved):
Looks good :star: 

One thing i noticed, but we could create a new issue:
When a view is maximized, we don't set it to active, which means we get a grey header if it's already active and a background-colored header if it's not active.

PR #919, opened 2024-11-28
feat(AssignmentRow): add click to open pen to edit

gularsson:
- Added Pen icon to AssignmentRow for editing
- Moved onClick event to AssignmentRowContent

### DIFF
--- a/src/views/Planning/components/AssignmentRow.tsx
+++ b/src/views/Planning/components/AssignmentRow.tsx
@@ -2,7 +2,7 @@ import { TimeDisplay } from '@/components/DataItem/TimeDisplay'
 import { AssignmentType } from '@/components/DataItem/AssignmentType'
 import { AssigneeAvatars } from '@/components/DataItem/AssigneeAvatars'
 import { DotDropdownMenu } from '@/components/ui/DotMenu'
-import { Delete, Edit, FileInput } from '@ttab/elephant-ui/icons'
+import { Delete, Edit, FileInput, Pen } from '@ttab/elephant-ui/icons'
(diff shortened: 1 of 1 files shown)


src/views/Planning/components/AssignmentRow.tsx
```diff
@@ -108,18 +101,22 @@ const AssignmentRowContent = ({ index, onSelect }: {
   }
 
   return (
-    <div className='flex flex-col gap-2 text-sm px-6 pt-2.5 pb-4 hover:bg-muted'>
+    <div
+      onClick={(event) => {
+        if (assignmentType === 'text' || assignmentType === 'flash') {
+          onOpenArticleEvent(event)
+        } else {
```
dannelundqvist, line 108: It might seem like a good idea to support editing photo assignments on click while opening text assignments article for editing. But it feels a bit confusing that clicking the row edit the photo assignment whereas clicking the text assignment row open the article for editing instead.

I think having a consistent behavior, even though nothing then happens when clicking the photo assignment row, is less confusing. At least I think we should discuss it more.

dannelundqvist (commented):
Looks good! See specific comment for discussion.

src/views/Planning/components/AssignmentRow.tsx
gularsson: Me and @hugowetterberg had that discussion yesterday. Yes, different outcome for the same action, however we both felt that it's the "primary" action for each, open vs edit.

dannelundqvist (approved):

PR #926, opened 2024-12-02
Feature/ele 935 wires view on smaller screens or with many wires

gularsson:


Victor-Lindh (approved):

gularsson:
> It would be nice to introduce something like "MultiView" so that we don't need to check for Wires in AppContent.tsx. But that could be done separately from this PR.

My thought as well, we would need some way to identify it as a MultiView, thats done a bit simplistic now (but works) for wires.

PR #941, opened 2024-12-04
Spellchecker should use document language not locale

Victor-Lindh:


src/defaults/indexedDBSpecification.ts
gularsson, line 1: You need to update the version for the indexedDBSpecification

gularsson (changes requested):
Nice :+1:

When opening a wire with `nb-no` or `da-dk` I'm getting this error:

```
{"code":"internal","msg":"failed to marshal proto response: string field contains invalid UTF-8","meta":{"cause":"*spell.wrappedError"}}
```

I don't know if it's related to the changes, but I can't remember getting it before.

src/defaults/indexedDBSpecification.ts
Victor-Lindh: fixed

Victor-Lindh:
> Nice 👍
> 
> When opening a wire with `nb-no` or `da-dk` I'm getting this error:
> 
> ```
> {"code":"internal","msg":"failed to marshal proto response: string field contains invalid UTF-8","meta":{"cause":"*spell.wrappedError"}}
> ```
> 
> I don't know if it's related to the changes, but I can't remember getting it before.

I'm looking into it.

src/datastore/contexts/SupportedLanguagesProvider.tsx
dannelundqvist, line 1: For readability I'd suggest to use early returns and more whitespace.

Instead of
```
   if (!session?.accessToken || !spellcheckUrl || !IDB.db) {
      return
    }
    const cachedLanguages: IDBLanguage[] | undefined = await IDB.get('languages')
    if (Array.isArray(cachedLanguages) && cachedLanguages.length) {
      setLanguages(cachedLanguages)
    } else {
       const supportedLanguagesUrl = ...
```

you could do

```
   if (!session?.accessToken || !spellcheckUrl || !IDB.db) {
      return
    }
    
    const cachedLanguages: IDBLanguage[] | undefined = await IDB.get('languages')
    
    if (Array.isArray(cachedLanguages) && cachedLanguages.length) {
      setLanguages(cachedLanguages)
      return
    }
    
    const supportedLanguagesUrl = ...
```

src/datastore/contexts/SupportedLanguagesProvider.tsx
```diff
@@ -0,0 +1,69 @@
...
+      if (response.ok) {
+        const result: unknown = await response.json()
+        if (typeof result === 'object'
+          && result !== null
+          && 'languages' in result
+          && Array.isArray(result.languages)) {
+          setLanguages(result.languages)
+          result.languages.forEach((lang: IDBLanguage, index: number) => {
+            void IDB.put('languages', {
+              id: `supported_language_${index + 1}`,
```
dannelundqvist, line 51: The id used does not help us in the future to retrieve or remove a specific language. Maybe we should consider having the language code as id? Looks a bit awkward now but allows us to expand on the data model later if needed or add functionality like adding languages through SSE or whatnot.

dannelundqvist (changes requested):

src/datastore/contexts/SupportedLanguagesProvider.tsx
Victor-Lindh: Added new commit to fix this!

src/datastore/contexts/SupportedLanguagesProvider.tsx
Victor-Lindh: Fixed in new commit

Victor-Lindh:
> > Nice 👍
> > When opening a wire with `nb-no` or `da-dk` I'm getting this error:
> > ```
> > {"code":"internal","msg":"failed to marshal proto response: string field contains invalid UTF-8","meta":{"cause":"*spell.wrappedError"}}
> > ```
> > I don't know if it's related to the changes, but I can't remember getting it before.
> 
> I'm looking into it.

Backend-related, @hugowetterberg is aware of the issue.

gularsson (approved):

src/views/Flash/FlashEditor.tsx
```diff
@@ -84,7 +87,10 @@ function EditorContent({ provider, user, setTitle }: {
     <Textbit.Editable
       yjsEditor={yjsEditor}
       onSpellcheck={async (texts) => {
-        return await spellchecker?.check(texts, locale, session?.accessToken ?? '') ?? []
```
dannelundqvist, line 87: The last `?? []` should be unnecessary as `.check()` always return `[]`

src/views/Flash/FlashEditor.tsx
```diff
@@ -84,7 +87,10 @@ function EditorContent({ provider, user, setTitle }: {
...
+        if (documentLanguage) {
+          return await spellchecker?.check(texts, documentLanguage, supportedLanguages, session?.accessToken ?? '') ?? []
```
dannelundqvist, line 91: The last `?? []` should be unnecessary as `.check()` always return `[]`

dannelundqvist (approved):
Nice work! Added two not so important comments.

src/views/Flash/FlashEditor.tsx
Victor-Lindh: Removing it makes TS emit an error. I'll see if I can make it work.

Victor-Lindh: Pushed new commit to mitigate the error. A bit more verbose now maybe, but maybe clearer?

src/views/Flash/FlashEditor.tsx
Victor-Lindh: Pushed new commit to mitigate the error. A bit more verbose now maybe, but maybe clearer?

PR #945, opened 2024-12-05
fix(snapshot): ignore userTracker in snapshot

gularsson:


dannelundqvist (approved):

Victor-Lindh (approved):
Looks good! What implications are there when we do snapshot `userTracker`?

gularsson:
> Looks good! What implications are there when we do snapshot `userTracker`?

We're trying to save it to the repository, which doesn't work since it's not a repo doc, and error occurs. It's a document of a more ephemeral nature.

PR #951, opened 2024-12-06
fix: focusable navigation create dialog

gularsson:


dannelundqvist (approved):

Victor-Lindh (commented):
Nice. Much clearer now!

Perhaps a separate issue, but should the slugs also be in black, in an opened planning item?
Skärmavbild 2024-12-06 kl  14 33 42

gularsson:
> Nice. Much clearer now!
> 
> Perhaps already a separate issue, but should the slugs also be in black, in an opened planning item? Skärmavbild 2024-12-06 kl 14 33 42

I think, in this case, we have yet another state for the slug. The one in the example is disabled. If it's changeable, it _is_ black.

PR #957, opened 2024-12-09
chore: remove deprecated functions

gularsson:
- Deleted `useYObserver` hook and its references
- Removed `yMapValueByPath` utility and its usage


dannelundqvist (approved):
Nice to have that cleaned up!

PR #958, opened 2024-12-09
fix(editor): focused fixes, menus and footer

gularsson:
- Added new Footer component to View
- Updated ContentMenu and ContextMenu with z-index


dannelundqvist (approved):
👍

PR #959, opened 2024-12-10
chore: remove cookie-parser dependency

gularsson:
I'm pretty sure this isn't used...


dannelundqvist (approved):
I'm pretty sure you're right.

PR #970, opened 2024-12-16
feat(handlelink): add keepFocus

gularsson:
- Added keepFocus boolean to LinkClick interface and handleLink function
- Updated Table component to pass keepFocus based on space key event
- Ensured viewId is retained if keepFocus is true


dannelundqvist (approved):
Nice!
It is good to get rid of the mention of Wires view in the generic table component.

PR #971, opened 2024-12-16
Feature/add table transform and revert

Victor-Lindh:


gularsson (commented):
Could we do a revertTable as well?

Victor-Lindh:
> Could we do a revertTable as well?

Added new commit to add a `revertTable` function.

PR #974, opened 2024-12-17
Feature/ele 850 approvals view

dannelundqvist:
* Basic approvals view with key navigation changes to accomodate key nav conflicts.
* Basic (largely unstyled) card component (that should be moved to elephant-ui later)
* Clock icon component that show the right clock icon based on time
* Search hook pattern that uses Twirp to access index instead of directly going towards openSearch (useAssignments)

Not in this PR:
* Still some issues with column widths when opening up article or planning view on the side.
* No filtering of text assignments based on status (or no/empty article)
* Drag'n drop or moving or reacting on sse


package.json
gularsson, line 1: Haven't noticed before that `@ttab/elephant-api` is in both devdeps and deps. Deps should be enough.

shared/Index.ts
```diff
@@ -0,0 +1,93 @@
...
+        errorCode: -1,
+        errorMessage: `Unable to query index: ${(err as Error)?.message || 'Unknown error'}`,
+        total: 0,
+        page: 0,
+        pageSize: 0,
+        pages: 0,
+        hits: []
+      }
+    }
+  }
```
gularsson, line 92: Could we provide the error code received from twirp?

gularsson (approved):

src/views/Approvals/index.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+                    <Card.Content>
+                      <Card.Title>{assignment.title}</Card.Title>
+                      <Card.Body>{assignment.meta.find((m) => m.type === 'tt/slugline')?.value || '-'}</Card.Body>
+                    </Card.Content>
+
+                    <Card.Footer>
+                      {assignment._section}
+                      &middot;
+                      Anders Andersson/TT
```
Victor-Lindh, line 150: Should we not type out the assignee, if found, here instead of a placeholder?

Victor-Lindh (approved):

src/views/Approvals/index.tsx
dannelundqvist: Definitely! Forgot to fix that. I'll get to it in follow up issues.

PR #985, opened 2024-12-20
feat(Approvals): space and enter actions, dotmenu

gularsson:


dannelundqvist (changes requested):
Pressing `space` when focus is on three dot menu should not open article preview.

dannelundqvist (approved):
Really nice work!

PR #996, opened 2025-01-02
Header title plugin should display character count

Victor-Lindh:


dannelundqvist (changes requested):
Requirements is that only the flash editor should display the character count. Not the textboxes. But I think we should keep it in the article editor for future discussion.

src/views/Editor/index.tsx
```diff
@@ -105,6 +105,9 @@ function EditorWrapper(props: ViewProps & {
         plugins={
           [
             ...plugins.map((initPlugin) => initPlugin()),
+            Text({
+              countCharacters: true
```
dannelundqvist, line 109: This must be wrong after the changes? Probably only the heading-1 should be used.

dannelundqvist (changes requested):

src/views/Editor/index.tsx
Victor-Lindh: Yes, a leftover. I'll remove it.

PR #1001, opened 2025-01-03
Approvals view should allow choosing date

Victor-Lindh:


src/views/Approvals/index.tsx
```diff
@@ -51,15 +56,24 @@ export const Approvals = (): JSX.Element => {
...
     type: 'text',
-    date: new Date(),
+    date: from ? new Date(from) : new Date(),
     slots,
     statuses: ['draft', 'done', 'approved', 'withheld']
   })
 
   const [focusedColumn, setFocusedColumn] = useState<number>()
   const [focusedCard, setFocusedCard] = useState<number>()
+  const [currentTab, setCurrentTab] = useState<string>('list')
```
dannelundqvist, line 76: As approvals view is currently only/always a grid, setting it as 'grid' would probably be more in line?

src/components/Header/Datechanger.tsx
```diff
@@ -19,7 +19,7 @@ export const DateChanger = ({ type }: {
 
   const steps = to ? 7 : 1
 
-  const validViews: View[] = ['Plannings', 'Events', 'Assignments']
+  const validViews: View[] = ['Plannings', 'Events', 'Assignments', 'Approvals']
```
dannelundqvist, line 22: I can see that this was already existing logic, but I don't see that it would be necessary for the date changer component to validate the user, maybe we can remove this check? What do you say @gularsson ?

dannelundqvist (changes requested):

src/views/Approvals/index.tsx
Victor-Lindh: Added new commit to fix this

src/components/Header/Datechanger.tsx
gularsson: This is a bit construed, and I barely remembered why...
It's because we're passing `type='Planning'` to the Header component which is needed in `CreateDocumentDialog` but in the DateChanger vi need to open the view `Plannings`.

TLDR; It's to make sure that we open Plannings, Events, Assignments, Approvals, even if we pass Planning, Assignment, Approval to the Header.

dannelundqvist: Ok, understand. Not perfect but nothing we should change as part of this PR. I'm letting it go for now :)

dannelundqvist (approved):

PR #1020, opened 2025-01-13
Feature/ele 974 approvals view should update automatically based on

dannelundqvist:


src/contexts/RepositoryEventsProvider.tsx
```diff
@@ -7,6 +7,7 @@ import { useIndexedDB } from '../datastore/hooks/useIndexedDB'
 export interface ElephantRepositoryEvent {
   event: string
   id: string
+  main_document: string
```
gularsson, line 10: I think we could use `import { EventlogItem } from '@ttab/elephant-api/repository'` instead.

gularsson (approved):
:+1:

src/contexts/RepositoryEventsProvider.tsx
dannelundqvist: Using the correct type also revealed a bug in the code. Good pointer!

PR #1021, opened 2025-01-13
feat(useAssignments): separate fetch, filter, and structure

gularsson:
- Moved fetch logic to `fetch.ts` to handle assignment fetching
- Added `filter.ts` to filter assignments based on statuses
- Created `structure.ts` to categorize assignments into time slots
- Updated `useAssignments.tsx` to use the new modular functions


dannelundqvist (approved):
Nice. Better structure now as well.

PR #1024, opened 2025-01-14
bugfix: apply new going backwards logic

Victor-Lindh:


gularsson (changes requested):
After discussion: we need to do history.go(delta) before for this to work (or similar, if possible), otherwise previous navigations are still in the history.

PR #1027, opened 2025-01-15
Feature/ele 998 approvals view should show active users in articles

dannelundqvist:
The change in CollaborationServer.tsx:295... might need more changes, not sure this works correct in a multi user env (hard to test locally).


gularsson (approved):
Looking good :star:

Victor-Lindh (approved):

gularsson (approved):

PR #1039, opened 2025-01-20
Removed image from stateless zod as it is not there and crash event c…

dannelundqvist:
…reation


hugowetterberg:
This updates the schema, but shouldn't we stop crashing on validation errors as well?

dannelundqvist:
> This updates the schema, but shouldn't we stop crashing on validation errors as well?

Yes, definitely, but as a separate fix.

Victor-Lindh (approved):

PR #1065, opened 2025-01-23
feat: use grid filter

gularsson:


src/components/GridFilter/Items/SectionFilter.tsx
```diff
@@ -0,0 +1,72 @@
...
+  })
+
+  if (!page) {
+    return (
+      <CommandItem
+        onSelect={() => setPage('section')}
+        className='flex gap-1 items-center'
+      >
+        <Shapes size={18} strokeWidth={1.75} />
+        Section
```
Victor-Lindh, line 28: The label here should probably use the swedish "Sektion"

src/components/GridFilter/Items/SectionFilter.tsx
```diff
@@ -0,0 +1,72 @@
...
+      <>
+        {allSections.map((section) => {
+          const isSelected = sections?.has?.(section.label)
+
+          return (
+            <CommandItem
+              className='flex gap-1 items-center'
+              key={section.value}
+              onSelect={() => {
+                console.log('onSelect')
```
Victor-Lindh, line 44: Leftover console.log

src/components/GridFilter/Items/StatusFilter.tsx
```diff
@@ -0,0 +1,69 @@
...
+      <>
+        {DocumentStatuses.map((status) => {
+          const isSelected = statuses?.has?.(status.value)
+
+          return (
+            <CommandItem
+              className='flex gap-1 items-center'
+              key={status.value}
+              onSelect={() => {
+                console.log('onSelect')
```
Victor-Lindh, line 38: Leftover console.log

src/components/GridFilter/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+export const GridFilter = (): JSX.Element => {
+  const [open, setOpen] = useState(false)
+  const [page, setPage] = useState<string>('')
+
+  const onOpenChange = useMemo(
+    () => handleOpenChange({ setOpen }),
+    [setOpen])
+
+  return (
+    <Popover open={open} onOpenChange={onOpenChange}>
```
Victor-Lindh, line 17: When I select one status filter, for example, the popover closes immediately. We could add modal={true} to the Popover component here to keep it open between selects.

src/components/GridFilter/Items/StatusFilter.tsx
```diff
@@ -0,0 +1,69 @@
...
+                  'mr-2 flex h-4 w-4 items-center justify-center rounded-sm border border-primary',
+                  isSelected
+                    ? 'bg-primary text-primary-foreground'
+                    : 'opacity-50 [&_svg]:invisible'
+                )}
+              >
+                <CheckIcon size={18} strokeWidth={1.75} />
+              </div>
+              {status.icon && (
+                <status.icon size={18} strokeWidth={1.75} />
```
Victor-Lindh, line 59: { ...status.iconProps } here to get the colors, maybe?

Victor-Lindh (changes requested):
If I open the Filter popover, then select "Status", there is no way to get back to the start of the Filter, meaning I can't get get to the "Section" filter here.

src/components/GridFilter/Items/StatusFilter.tsx
gularsson: Hmm, maybe, but we're not having colors in Planning or Events.

Victor-Lindh: That's true, except for here in this view:
Skärmavbild 2025-01-27 kl  10 19 33

Separate issue, in any way.

dannelundqvist (changes requested):
When you have chosen to filter on section (or the other) it is difficult (impossible?) to get back and filter on section.

https://github.com/user-attachments/assets/fbc838a6-d585-4903-a5bf-df5884472387

dannelundqvist:
Another issue is probably not the fault of this PR, but the filter select box automatically closes every now and then, which can be quite annoying. It might ber that SSE event triggers a rerender. Not sure if this should be fixed in this change or somewhere else.

Victor-Lindh (changes requested):
The "Kultur & nöje" filter doesn't return any hits, even when there are cards with that section in one of the columns

Victor-Lindh (approved):

dannelundqvist (approved):

PR #1069, opened 2025-01-24
Assignment display exact times

Victor-Lindh:


dannelundqvist:
Skärmavbild 2025-01-24 kl  11 36 51

Really nice with the icon and the tooltip. It would be nice to not add the tooltip when it is empty (full day/heldag).

Additionally the time icon is visually paired with the assignment type rather than the time text which makes it a bit odd. They need to be visually "together", the time text and the icon. Maybe even switch places?

dannelundqvist (changes requested):
Works really good! Only issue with the title/tooltip and a discussion around placement/margin between time and icon.

Victor-Lindh:
> Works really good! Only issue with the title/tooltip and a discussion around placement/margin between time and icon.

I agree that the relation between the icon and what column it belongs to becomes a bit unclear here.
I'll make a change to the placement.

dannelundqvist (approved):

PR #1079, opened 2025-01-28
Feature/display whom changed status to done

Victor-Lindh:


gularsson (changes requested):
I seem to have missed to bump the IndexedDB version in my previous commit. Which caused a crash due to `sub` being undefined.

Could you please bump the version in this pull request?
Maybe we also need to prevent the crash with some error handling.

gularsson (approved):

PR #1080, opened 2025-01-28
compare user sub with author sub instead of email

Victor-Lindh:


src/views/Assignments/index.tsx
```diff
@@ -46,8 +46,11 @@ export const Assignments = (): JSX.Element => {
   const { data: session } = useSession()
 
   const assigneeUserName = useMemo(() => {
-    const userEmail = session?.user.sub
-    const author = authors.find((a: IDBAuthor) => a?.email === userEmail)
+    const userSub = session?.user?.sub
+    const subId = userSub?.slice(userSub?.lastIndexOf('/') + 1)
```
dannelundqvist, line 50: Is not `session.user.sub` just a uuid? If so, this line is not doing anything.

dannelundqvist (approved):

src/views/Assignments/index.tsx
dannelundqvist: Seems it was my user that was the odd setup as having a uuid as sub.

PR #1084, opened 2025-01-29
fix: validate flash, no newsvalue, no author

gularsson:
Some misunderstandings straightened out.
Flash shouldn't have: newsvalue, since it's implied, neither should it have an author (it's on the assignment)


dannelundqvist (approved):
Works well event without an author document. 👍

PR #1086, opened 2025-01-29
Calendar lagg till planering doesnt add details from the event as defaults

Victor-Lindh:


dannelundqvist (commented):
It works well both for existing event and when creating a new event with a planning inline. Nice!

Asked for clarification in the internal issue regarding desription and date/time which is currently not being copied. I think we should be certain before finishing this one.

dannelundqvist (approved):

PR #1100, opened 2025-01-31
feat: create, edit and delete an editorial note

gularsson:


src/views/Editor/components/Notes/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+          >
+            {selectedOptions?.[0] && SelectedIcon && (
+              <div className='flex'>
+                <SelectedIcon {...selectedOptions[0].iconProps} />
+              </div>
+            )}
+          </ComboBox>
+          <TextBox
+            path={`meta.core/note[${noteIndex}].data.text`}
+            placeholder='Add a note'
```
Victor-Lindh, line 58: I think the placeholder should be in Swedish.
"Lägg till en [intern/publik] beskrivning", or something like that.

src/views/Editor/components/Notes/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+          </Button>
+        </div>
+      </div>
+    </Alert>
+  )
+}
+
+export const Notes = (): JSX.Element | null => {
+  const [notes, setNotes] = useYValue<Block[] | undefined>('meta.core/note')
+  console.log(notes)
```
Victor-Lindh, line 78: Tidying up 🧹

src/views/Editor/components/Notes/index.tsx
Victor-Lindh: By the way, is "notering" the phrasing we're going with?
I think they are use "anteckning" in DAWROC today.
I'm fine with either, just wanted to put the discussion on the table 😄

Victor-Lindh:
Works great, just a couple of comments about forgotten console log and some phrasing.

src/views/Editor/components/Notes/index.tsx
gularsson: I like "notering" better, I'll stick with that until somebody want to change it ;)

dannelundqvist (approved):

PR #1102, opened 2025-01-31
Event date picker should now allow end before start time

Victor-Lindh:


gularsson (approved):
:+1: 

Improvment suggestion, we could use the <Validation /> component as well to indicate which input has an incorrect value.

gularsson (approved):

dannelundqvist (approved):

PR #1105, opened 2025-02-03
Feature/view document source

dannelundqvist:
Extremely simple view document source function  using `CMD/CTRL` + `SHIFT` + `u`

### DIFF
--- a/src/contexts/CollaborationProvider.tsx
+++ b/src/contexts/CollaborationProvider.tsx
@@ -12,6 +12,9 @@ import { Collaboration } from '@/defaults'
 import { HPWebSocketProviderContext } from '.'
 import type * as Y from 'yjs'
 import { createStateless, StatelessType } from '@/shared/stateless'
+import { useModal } from '@/components/Modal/useModal'
+import { useKeydownGlobal } from '@/hooks/useKeydownGlobal'
+import { Button } from '@ttab/elephant-ui'
 
 export interface AwarenessUserData {
   name: string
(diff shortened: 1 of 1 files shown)


src/contexts/CollaborationProvider.tsx
```diff
@@ -65,6 +68,30 @@ export const CollaborationProviderContext = ({ documentId, document, children }:
   const [connected, setConnected] = useState<boolean>(false)
   const [provider, setProvider] = useState<HocuspocusProvider>()
 
+  // Developer tool to display current document source in a dialog
+  const { showModal, hideModal } = useModal()
+  useKeydownGlobal((evt) => {
+    if (evt.key === 'u' && evt.shiftKey && evt.metaKey) {
```
gularsson, line 74: I need this for it to work, don't know if it's OS specific
```suggestion
    if (evt.key === 'U' && evt.shiftKey && evt.metaKey) {
```

src/contexts/CollaborationProvider.tsx
```diff
@@ -65,6 +68,30 @@ export const CollaborationProviderContext = ({ documentId, document, children }:
...
+        <div className='whitespace-pre flex flex-col gap-2 h-full overflow-hidden p-2'>
+          <h1 className='text-lg font-bold shrink p-2'>Document source</h1>
+
+          <div className='text-xs font-mono overflow-auto p-2 border bg-muted rounded-md'>
+            {JSON.stringify(provider.document.toJSON(), null, 2)}
+          </div>
+
+          <div className='shrink text-right p-2'>
+            <Button onClick={hideModal}>Close</Button>
+          </div>
```
gularsson, line 89: Maybe a Header with a X to close, it was a bit long and the Close-button (Which maybe should be "Stäng"?) ended up outside of viewport and i had to scroll.

gularsson (changes requested):

src/contexts/CollaborationProvider.tsx
gularsson: and just a note, metaKey = windows key for me.

Victor-Lindh: I can't seem to make it work at all.

src/contexts/CollaborationProvider.tsx
```diff
@@ -65,6 +68,30 @@ export const CollaborationProviderContext = ({ documentId, document, children }:
...
+      if (!provider?.document) {
+        return
+      }
+
+      showModal(
+        <div className='whitespace-pre flex flex-col gap-2 h-full overflow-hidden p-2'>
+          <h1 className='text-lg font-bold shrink p-2'>Document source</h1>
+
+          <div className='text-xs font-mono overflow-auto p-2 border bg-muted rounded-md'>
+            {JSON.stringify(provider.document.toJSON(), null, 2)}
```
Victor-Lindh, line 84: Don't know if it's just me, but I'm getting a deprecation warning through vsCode:
`@deprecated — Do not use this method and rather call toJSON directly on the shared types`

Seems to be solved by doing this:
`{JSON.stringify(provider.document.getMap('ele').toJSON(), null, 2)}`

dannelundqvist: Not just you. Nice catch!

Victor-Lindh (approved):
Nice work!

gularsson (approved):

PR #1112, opened 2025-02-04
fine-control popover content render position to prevent it from poppi…

Victor-Lindh:
…ng up outside of the viewport


gularsson (approved):
Works fine on desktop, however it's not really mobile friendly:
Screenshot_2025-02-04-10-37-36_314

How do you feel about implementing something similar that we have for the ComboBox, if we're on mobileish screen size then we'll use a drawer instead?

I'll approved cause this makes it way better than before :slightly_smiling_face:

PR #1113, opened 2025-02-04
fix: keep empty cells, but add empty string to content

gularsson:


src-srv/utils/transformations/lib/parseTableData.ts
```diff
@@ -19,13 +19,13 @@ export function parseTableBody(tablebody: string) {
...
       class: 'block',
-      children: cells.filter((cell) => {
-        return cell.textContent.trim().length > 0
-      }).map((cell) => {
+      children: cells.map((cell) => {
         return {
           type: 'core/table/row/cell',
           class: 'text',
-          children: [{ text: cell.textContent.trim() }]
+          children: cell.textContent.trim().length > 0
```
dannelundqvist, line 26: Is this not a bit overly complex? Instead of
```
cell.textContent.trim().length > 0
   ? [{ text: cell.textContent.trim() }]
   : [{ text: '' }]
```

it should be enough with just

```
[{ text: cell.textContent.trim() }]
```

dannelundqvist (changes requested):

dannelundqvist (approved):

PR #1114, opened 2025-02-04
Feature/ele 992 inline element not implemented strong

dannelundqvist:
Added support for strong, em and refactored transoformation of newsdoc to slate to newsdoc


gularsson (approved):
:+1:

Victor-Lindh (approved):

PR #1120, opened 2025-02-05
feat: add toggle and toggle-group

gularsson:
Depends on: https://github.com/ttab/elephant-ui/pull/595


gularsson:
> Looks great! Although, it doesn't seem to instantly mark the row when either of the Toggles are clicked, or rather, the UI is not reflecting the selection until a hard refresh.

Yes, I know, unfortunately. I was planning on doing the event listening/mutate in a separate PR. Since that is kinda separated from this feature.

Victor-Lindh:
Skärmavbild 2025-02-06 kl  11 09 26
During testing we discovered this rerendering bug, Gustav has a solution incoming.

Victor-Lindh (approved):

PR #1121, opened 2025-02-05
fix: check wires data better

gularsson:


src/views/Wires/components/Controller.tsx
```diff
@@ -12,8 +12,10 @@ export const Controller = (): JSX.Element => {
 
   useEffect(() => {
     // Load state from userTracker if it doesn't match the current state
-    if (wiresHistory && wiresHistory.contentState[0].viewId !== state?.contentState[0].viewId) {
-      replaceState(wiresHistory?.contentState[0].path, wiresHistory)
+    if (wiresHistory && wiresHistory.contentState
+      ?.every((contentState, index) =>
```
danijelvukoje, line 16: This should be 
```suggestion
      ?.some((contentState, index) =>
```
to make sure that we replace the state if any of the views is not same as in userTracker state.

danijelvukoje (changes requested):

src/views/Wires/components/Controller.tsx
gularsson: Great insight :+1:

danijelvukoje (approved):
Looks great!

Victor-Lindh (approved):

PR #1128, opened 2025-02-07
feature: show published articles

Victor-Lindh:


dannelundqvist (changes requested):
We should move the menu option to above the separator line. The main menu should close when opening the published articles sheet.

dannelundqvist (approved):

PR #1134, opened 2025-02-11
feat: use events to update wires views

gularsson:


dannelundqvist (approved):
Works really well. I like the "short" names of "fetch", "handlers" etc. We _must_ have the same naming convention for all these large hooks. And I think we really need to make assignments follow the same clear naming convention. (But as a separate PR).

Having an index.tsx barrel file is not needed. I would event say it pollutes the project with unnecessary files, hides the file structure when importing and slows down build and test times. (We should get rid of many of them...) But I won't stand on the barricades... ;)

Good work!

Victor-Lindh (approved):

PR #1149, opened 2025-02-19
Feature/ele 1113 create text assignment with deliverable from a wire

gularsson:
Known TODOS:
* ~Time, flash, article, default values in template~
* Sluglines, we need to determine that we don't have duplicate sluglines, and create unique sluglines (probably user input when creating an article from a wire, only with existing planning)
*More template payloads could be harmonized.
* I broke the function to transfer data from event to planning, creating. Neeed to be redone.
* ~Searches with Twirp are _not_ case sensitive~
* Searches with Twirp should be mulit_match instead of prefix


src/views/Wire/WireViewContent.tsx
```diff
@@ -0,0 +1,213 @@
...
+          && <ViewHeader.RemoteUsers documentId={props.id} />}
+        </ViewHeader.Action>
+      </ViewHeader.Root>
+
+      <View.Content>
+        <Form.Root asDialog={props.asDialog}>
+          <Form.Content>
+            <Form.Group icon={Cable}>
+              <div className='w-full'>
+                <Input
```
Victor-Lindh, line 62: Skärmavbild 2025-02-20 kl  13 48 06
If the headline is long, it will appear outside of the Dialog window. It's scrollable on the x-axis though.
We do have the `Textarea` component in `elephant-ui` that we could utilize here, if we want to make sure the full headline is visible.

src/views/Wire/WireViewContent.tsx
```diff
@@ -0,0 +1,213 @@
...
+                  className='pl-0 pt-2 h-8 text-medium border-0'
+                  readOnly
+                  value={props.wire.fields['document.title'].values?.[0]}
+                />
+              </div>
+            </Form.Group>
+            <Form.Group icon={GanttChartSquare}>
+              <Awareness name='FlashPlanningItem' ref={planningAwareness}>
```
Victor-Lindh, line 70: `FlashPlanningItem` should be changed to `WirePlanningItem`, maybe?

Victor-Lindh:
After creating an article I noticed that assignee is not visible in the assignment list (in the planning item), nor is the newsvalue.

Skärmavbild 2025-02-20 kl  14 37 20

src/views/Wire/WireViewContent.tsx
```diff
@@ -0,0 +1,213 @@
...
+                  }
+                }}
+              />
+            )
+          }
+
+          <Form.Footer>
+            <Form.Submit onSubmit={handleSubmit}>
+              <div className='flex justify-end'>
+                <Button type='submit'>Skapa artikel från telegram</Button>
```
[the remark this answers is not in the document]
gularsson: Yeah, i felt that too, however i opted for clarity.

gularsson:
> After creating an article I noticed that assignee is not visible in the assignment list (in the planning item), nor is the slug.
Hmm, yeah, slug should be there. I'll fix. However I'm not sure that we should automatically use the creator as assignee. So maybe it's best left empty?

gularsson:
> After creating an article I noticed that assignee is not visible in the assignment list (in the planning item), nor is the slug.
Hmm, yeah, slug should be there. I'll fix. However I'm not sure that we should automatically use the creator as assignee. So maybe it's best left empty?

gularsson:
> After creating an article I noticed that assignee is not visible in the assignment list (in the planning item), nor is the slug.
> Skärmavbild 2025-02-20 kl 14 37 20
> 
> compared to the usual look:
> Skärmavbild 2025-02-20 kl 14 54 08

Is this when creating a new planning or using an existing?

Victor-Lindh:
1. Super small (but annoying) is that the double quotation marks around the planning name (Techarena dag1) have no closing mark.
Skärmavbild 2025-02-21 kl  08 57 51

2. When I create an article from a wire, and select a planning in which the article should be listed in the assignments list, it doesn't seem to be picked up? 🤔 
It's visible in the Planning Overview, though 👍🏻
 
Skärmavbild 2025-02-21 kl  09 01 13

Skärmavbild 2025-02-21 kl  09 02 00

gularsson:
> 1. Super small (but annoying) is that the double quotation marks around the planning name (Techarena dag1) have no closing mark. Can wait until later.
> 
> 
> Skärmavbild 2025-02-21 kl 08 57 51
> 
>     2. When I create an article from a wire, and select a planning in which the article should be listed in the assignments list, it doesn't seem to be picked up? 🤔
>        It's visible in the Planning Overview, though 👍🏻
> 
> 
> Skärmavbild 2025-02-21 kl 09 01 13 Skärmavbild 2025-02-21 kl 09 02 00

1. Will fix
2. I regard this as an edge case with faulty imported data, we dont have a slugline, so we can not persist the planning, hence we're not able to add assignment or article.

src/views/Wire/WireViewContent.tsx
dannelundqvist: But it would be nice to use ellipsis for the truncation.

dannelundqvist (approved):

PR #1159, opened 2025-02-21
fix: use multi_match query again, but with Twirp

gularsson:


Victor-Lindh (changes requested):
I'm having problems getting wanted results back when searching in the Flash view:
Searching for a planning in the list in the background, which is clearly there.
No results.

Trying the same query but in the exact casing as the planning title.
No results.
Trying different combinations of multimatchquery options, getting varied results back.

Edit:
I've found that if you configure the query like this:
```
                multiMatch: MultiMatchQueryV1.create({
                  fields: ['document.title', 'document.rel.section.title'],
                  query,
                  type: 'phrase_prefix'
                })
```
, you get better results back.
However, searching for items that start with quotation marks doesn't seem to return correct result.

Victor-Lindh (approved):

PR #1160, opened 2025-02-21
fix: remove outline from dialogs from useModal

gularsson:


Victor-Lindh (commented):
Should the the same be done for other DialogContent in the application as well?

gularsson:
> Should the the same be done for other DialogContent in the application as well?

You're right, should probably be done in `@ttab/elephant-ui`

PR #1170, opened 2025-02-25
feat: mark newsvalue 6 wires as priority

gularsson:


Victor-Lindh (commented):
Should the priority be marked in some way inside the Preview as well?

Victor-Lindh (approved):

PR #1185, opened 2025-03-03
Feature: duplicate event

Victor-Lindh:


gularsson (commented):
As mentioned on the meeting we should probably add the functionality here:
Screenshot_2025-03-03-14-25-27_20752
And we should find another place for it in the view, not sure where though.

Händelse-menu could be clickable, we could add a tripple-dot menu to the header, or we could copy the sheet used in editor. Non is perfect though.

src/components/Duplicate/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+          provider={provider}
+          title={title || ''}
+          description={`Vill du kopiera ${type === 'event' ? 'händelsen' : ''} till ${format(duplicateDate, 'dd/MM/yyyy')}?`}
+          secondaryLabel='Avbryt'
+          primaryLabel='Kopiera'
+          onPrimary={(duplicateId: string | undefined) => {
+            console.log('provider', provider)
+            if (provider && status === 'authenticated' && duplicateId && session) {
+              try {
+                console.log('sending', duplicateId)
```
gularsson, line 62: :broom:

src/components/Duplicate/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+                    state: false,
+                    id: duplicateId,
+                    context: {
+                      accessToken: session.accessToken,
+                      user: session.user,
+                      type: type.slice(0, 1).toUpperCase().concat(type.slice(1))
+                    }
+                  })
+                )
+                toast.success('Kopieringen lyckades!')
```
gularsson, line 74: Would be nice with a link to the event here

src/defaults/templates/duplicateTemplate.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  const keys: Array<keyof EleDocument> = ['meta', 'links']
+
+  const doc: DocumentType = keys.reduce((newDoc, key) => {
+    if (!document || !newsDocDocument || !newsDocDocument[key]) {
+      return newDoc
+    }
+
+    const value: Record<string, BlockType[]> = newsDocDocument[key]
+    console.log('value', value)
```
gularsson, line 59: :broom:

src/defaults/templates/duplicateTemplate.ts
gularsson, line 1: Did you try cloning the initial document? I still think it would be more terse and less error prone.

gularsson (changes requested):

src/defaults/templates/duplicateTemplate.ts
Victor-Lindh: Did try, didn't get it to work

src/components/Duplicate/index.tsx
Victor-Lindh: Added new commit that does this

dannelundqvist:
> As mentioned on the meeting we should probably add the functionality here: ![Screenshot_2025-03-03-14-25-27_20752]
I would suggest we simply start with moving the button to the header (and maybe then adjust borders and background if needed to make it look nice). Then we can think harder on how to incorporate a dot menu or something else as a pattern for all views.

Skärmavbild 2025-03-03 kl  16 44 29

dannelundqvist:
I think we should assume things work out as we want it. Something like `Händelsen "Pressträff om ny generaldirektör" kopierades till 2025-03-12` would be more useful to the user and remind the user where the copy went.

When/if we later collect all notifications in a notification center/archive that text would also be more useful than five notifications that says "Kopiering lyckades!". We want to establish a pattern for notification texts.

dannelundqvist:
> I think we should assume things work out as we want it. Something like `Händelsen "Pressträff om ny generaldirektör" kopierades till 2025-03-12` would be more useful to the user and remind the user where the copy went.

src/defaults/templates/duplicateTemplate.ts
```diff
@@ -0,0 +1,103 @@
...
+
+export function duplicateTemplate(id: string, payload: { type: string, newDate: string, document: Doc }): DocumentType {
+  const { documentResponse: { document: newsDocDocument } } = fromYjsNewsDoc(payload?.document)
+
+  function mergeDateWithTime(date1ISO: string, date2ISO: string) {
+    if (!date1ISO || !date2ISO) {
+      return ''
+    }
+
+    const [_, time] = date1ISO.split('T')
```
dannelundqvist, line 26: Not that it matters, but there is no real reason to use _ for this case. It could be written as
```typescript
const [, time] = date1ISO.split('T')
```

Victor-Lindh: fixed in new commit

Victor-Lindh:
Added new commit that moves it into the view header. Also, replaced the button-esque trigger with a suitable icon instead, to save some space.

And we should find another place for it in the view, not sure where though.  Händelse-menu could be clickable, we could add a tripple-dot menu to the header, or we could copy the sheet used in editor. Non is perfect though.

gularsson (approved):

PR #1186, opened 2025-03-03
Fix: event list times

Victor-Lindh:


gularsson (approved):
Looks good, did a compare and they seem to be there now! :+1:

PR #1188, opened 2025-03-03
Add correct props to awareness and view header title

dannelundqvist:


src/views/Error/index.tsx
```diff
@@ -27,7 +27,7 @@ export const Error = ({ title, error, message }: ViewProps & {
   return (
     <View.Root>
       <ViewHeader.Root>
-        <ViewHeader.Title title='Ett fel har uppstått!' icon={MessageCircleWarning} />
+        <ViewHeader.Title name='Error' title='Ett fel har uppstått!' icon={MessageCircleWarning} />
```
gularsson, line 30: name is typed as string in `ViewHeader.Title` I would prefer:
```ApplicationMenuItem['name']``` to ensure that is a valid option

gularsson (approved):

PR #1190, opened 2025-03-04
feat: persist filters in userTrackerDocument

gularsson:
Feat: persist filters in userTracker
Fix: removed filters should be set to undefined when passed to query

We're not loading the saved filter when navigating to an unfiltered view that has a saved filter. It's a user action to select the saved filer.

It's a more transparent choice for the user, and saves us from a lot of juggling.
However it's prepared if we want to have it in the future. Some bugs and refactoring where made, and that's included in this PR.

One thing we might have to revisit is that we're now halting until Init is loaded, which doesn't seem to work on a hot reload. So now we have to refresh after a hotload, a bit of a nuisance, but it will guarantee that the userTracker works and will save us from a lot of unexplained issues.

### DIFF
--- a/src/components/GridFilter/Toolbar.tsx
+++ b/src/components/GridFilter/Toolbar.tsx
@@ -1,71 +1,139 @@
-import { Save, X } from '@ttab/elephant-ui/icons'
+import { Save, UserCog, X } from '@ttab/elephant-ui/icons'
 
 import { Button, ToggleGroup, ToggleGroupItem } from '@ttab/elephant-ui'
 import { SelectedFilters } from './SelectedFilters'
+import type { Filter } from '@/hooks/useFilter'
 import { useFilter } from '@/hooks/useFilter'
 import { useSections } from '@/hooks/useSections'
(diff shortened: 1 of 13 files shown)


src/components/GridFilter/Toolbar.tsx
```diff
@@ -1,54 +1,139 @@
...
+      : undefined
+  }
+
+  const handleToggleValueChange = (value: string | undefined) => {
+    if (value === 'user') {
+      setFilter(userFilters as Filter || {})
+      return
+    }
+
+    // TODO remove ''
```
dannelundqvist, line 53: If it is as simple as removing '' – should we not do it? Or write more info ;)

dannelundqvist (approved):

src/components/GridFilter/Toolbar.tsx
gularsson: Yes, it's a part of a parallel refactor, just a temporary note to myself. Should probably have been handled :slightly_frowning_face:

PR #1201, opened 2025-03-10
Feature/ele 1078 change status component improvements

dannelundqvist:
Status change for articles, events and plannings. Many places now use the workflow/status specification information. Filters still use the older documentstatuses/defaultValueOption spec.

Not part of this PR:
* Withheld/scheduling
* Exact workflow items - including titles and descriptions
* Issue regarding going "backwards" in statuses (e.g. from usable to draft)


src/defaults/workflowSpecification.tsx
```diff
@@ -0,0 +1,277 @@
...
+
+export type WorkflowSpecification = Record<string, WorkflowState>
+
+export interface StatusSpecification {
+  icon: LucideIcon
+  className: string
+}
+
+export const StatusSpecifications = {
+  draft: {
```
gularsson, line 30: Maybe
```suggestion
export const StatusSpecifications: Record<string, StatusSpecification> = {
```
instead of typecasting it?

gularsson (approved):

src/defaults/workflowSpecification.tsx
dannelundqvist: Better stick to the regular pattern. Agree!

PR #1205, opened 2025-03-12
feat: wires search params, refactor filter component

gularsson:


dannelundqvist:
Skärmavbild 2025-03-12 kl  16 17 01

We should use the same margin/padding as for the menu and the content in the filter.

dannelundqvist:
Skärmavbild 2025-03-12 kl  16 19 04

The "clear filter" button is higher than other elements in the filter bar. This has the result that all content is jumping up and down a few pixels whenever a filter is selected or deselected.

dannelundqvist:
Skärmavbild 2025-03-12 kl  16 42 23
Selecting filters using the menu in "Dagen" does not work. (It works in all other places though)

dannelundqvist:
Skärmavbild 2025-03-12 kl  16 54 28
Should the leftmost menu (where you can select other filters) disappear in smaller container/screen widths?

dannelundqvist (changes requested):
I have no remarks on the code. And overall this is huge improvement! Nice work! I have found two functional issues that might need looking into.

dannelundqvist (approved):
Really good work!

PR #1207, opened 2025-03-13
Fix toggle filter bugs

gularsson:


Victor-Lindh (approved):
https://github.com/user-attachments/assets/1f50e3aa-12fa-4a6e-abee-515e50b9b0ca

PR #1209, opened 2025-03-14
feat: add sorting and grouping on tables

gularsson:


Victor-Lindh (changes requested):
Grouping by Section in Assignments view shows the section uuid in the group header instead of the section label.

Skärmavbild 2025-03-17 kl  09 43 10

Victor-Lindh (commented):
Grouping on start time in Assignments view displays undefined in the grouping header.

Skärmavbild 2025-03-17 kl  11 29 52

Victor-Lindh (approved):
Awesome!

PR #1214, opened 2025-03-17
fix: assignment type should be editable until article is created

gularsson:


dannelundqvist:
It must not be possible to change from text (or flash) assignment when a deliverable exists.

https://github.com/user-attachments/assets/2fc2ea4f-cc47-4304-b1ab-e1e8c632857f

dannelundqvist (changes requested):
Seems it is still possible to change assignments with existing deliverables to something else. Effectively creating orphan articles/flashes.

dannelundqvist (approved):

PR #1215, opened 2025-03-17
fix: use select and file text icon

gularsson:
based from: https://github.com/ttab/elephant-chrome/pull/1214


dannelundqvist (approved):
Works like a charm!

PR #1217, opened 2025-03-17
fix: append current author as assignee when creating planning

gularsson:


Victor-Lindh (changes requested):
Created a planning, added an assignment (was auto-assigned 👍🏻 ), but after focusing away from the tab and then back again I'm hit with this validation error.

Skärmavbild 2025-03-17 kl  16 38 51

Victor-Lindh (approved):

PR #1218, opened 2025-03-17
User API integration into the backend

danijelvukoje:


gularsson (approved):
Looks good, some minor things to sooth my OCD. 
Could you also please add the necessary envs to .env.tpl

danijelvukoje:
Thank you for the feedback, implemented all the suggestions.

PR #1222, opened 2025-03-18
Removed unused fetched workflow data from backend

dannelundqvist:


shared/Workflow.ts
```diff
@@ -36,41 +34,18 @@ export class Workflow {
       return this.#workflows[type]
     }
 
-    const [workflow, statuses] = await Promise.all([
-      this.#fetchWorkflow({ type, accessToken }),
+    const [statuses] = await Promise.all([
+      // this.#fetchWorkflow({ type, accessToken }),
       this.#fetchStatuses({ type, accessToken })
     ])
```
gularsson, line 40: Might as well remove this?
```suggestion
    const statuses = await this.#fetchStatuses({ type, accessToken })
```

gularsson (changes requested):

gularsson (approved):

PR #1232, opened 2025-03-21
long poll for user messages and display them in toasts

danijelvukoje:


src/components/UserMessagesSender.tsx
```diff
@@ -0,0 +1,87 @@
...
+  return (
+    <>
+      {children}
+    </>
+  )
+}
+
+async function pollMessages(accessToken: string, user: User) {
+  let lastId = -1
+  do {
```
dannelundqvist, line 33: Does the lastId ever become -500? If not, you could do a simple loop as it would stop on an exception anyway.
```
while(true) {
  lastId = await execPolling(accessToken, user, lastId)
}
  ```

dannelundqvist (changes requested):
Very well structured! Great!

A minor question around the infinite loop. I would also like to see a name change on the component to something like `UserMessageReceiver` as it receives/polls for messages, it doesn't send them.

gularsson (approved):
:+1:

src/components/UserMessagesSender.tsx
danijelvukoje: It doesn't. I agree, that was unnecessary. Renamed the component and simplified the loop handling.

dannelundqvist (approved):

PR #1233, opened 2025-03-21
Fix assignment times and remove excessive Popover

Victor-Lindh:


dannelundqvist:
A lot cleaner. And it works very well. Unfortunately the time is still displayed wrong when not editing the assignment.
Skärmavbild 2025-03-24 kl  11 23 21
Skärmavbild 2025-03-24 kl  11 23 30

dannelundqvist (changes requested):
See general comment about the time in non editing mode of the assignment.

Victor-Lindh:
> A lot cleaner. And it works very well. Unfortunately the time is still displayed wrong when not editing the assignment.

The 05:45 timestamp is the assignment's _publish_ time, while the 01:01 timestamp is the end time of the assignment.
Perhaps we should always show the publish time when editing as well, if it's present?
And also make sure it's not editable if published?

dannelundqvist:
> > A lot cleaner. And it works very well. Unfortunately the time is still displayed wrong when not editing the assignment.
> 
> The 05:45 timestamp is the assignment's _publish_ time, while the 01:01 timestamp is the end time of the assignment. Perhaps we should always show the publish time when editing as well, if it's present? And also make sure it's not editable if published?

Should not text assignments publish, start and end time be the same?

Victor-Lindh:
> Should not text assignments publish, start and end time be the same?

Not sure, tbh. Should they?

dannelundqvist (approved):

PR #1242, opened 2025-03-26
feat: move assignment to existing or new planning

gularsson:


Victor-Lindh (changes requested):
I get some validation errors in certain cases when no target planning is selected, which I've DM:ed about.

Also, in the cases when validation fails (and therefore not saved to a new planning), the assignment is still removed from the original planning, which should not happen until confirmed success.

danijelvukoje:
Same here, seems like there is some validation errors when trying to snapshot plannings with assignments. Also if I try to move an assignment without choosing an existing planning to the same date, it "succeeds", error shows up in a toast and the assignment disappears.

danijelvukoje (approved):
Nice, it seems to be running smoothly now 👍

Victor-Lindh (approved):
👍🏻

PR #1244, opened 2025-03-27
fix: calculate when to refresh, and keep original expire

gularsson:


src-srv/utils/authConfig.ts
```diff
@@ -39,11 +39,14 @@ async function refreshAccessToken(token: JWTPayload): Promise<JWTPayload> {
...
 
-    const refreshedTokens = await response.json()
+    const refreshedTokens = await response.json() as {
+      access_token: string
+      expires_in: number
+      refresh_token?: string
+    }
+
 
     if (!response.ok) {
```
dannelundqvist, line 51: Checking the response status should be done before line 44 with `const refreshedTokens = await response.json() as...`

src-srv/utils/authConfig.ts
```diff
@@ -39,11 +39,14 @@ async function refreshAccessToken(token: JWTPayload): Promise<JWTPayload> {
         Accept: 'application/json'
       },
       body: params
-    }).catch((ex) => {
-      throw new Error('refresh token grant request', { cause: ex })
     })
 
-    const refreshedTokens = await response.json()
+    const refreshedTokens = await response.json() as {
```
dannelundqvist, line 44: This could be unsafe. It could be better to add a typeguard check.

```
const refreshedTokens = await response.json() as unknown
 if (!isRefreshedTokens(refreshedTokens)) {
     // error... using throw, return etc
 }
  
 // ...rest of code
 
 
 function isRefreshedTokens(value: unknown): value is JWTPayload & { expires_in: string} {
  return typeof value === 'object' && value !== null // Ensure it's a non-null object
    && 'access_token' in value && typeof value.access_token === 'string'
    && 'expires_in' in value && typeof value.expires_in === 'string'
    && 'refresh_token' in value && typeof value.refresh_token === 'string'
}
 ```

dannelundqvist (changes requested):
Seems to work really well!
But some error handling could be improved, at the least the response.ok check.

dannelundqvist (approved):

PR #1250, opened 2025-03-31
fix: retry with new token

gularsson:


src/contexts/RepositoryEventsProvider.tsx
```diff
@@ -176,10 +188,25 @@ export const RepositoryEventsProvider = ({ children }: {
...
-      void listenForEvents(repositoryEventsUrl, data.accessToken)
-    }, 0)
-  }, [hasFocus, listenForEvents, repositoryEventsUrl, data?.accessToken])
+    let isRetryScheduled = false
+
+    const retryListenForEvents = () => {
+      const accessToken = tokenRef.current || ''
+      void listenForEvents(repositoryEventsUrl, accessToken).catch((ex) => {
+        if (ex instanceof AuthenticationError && !isRetryScheduled) {
+          console.info('Retrying listening for events in 20 seconds...')
```
danijelvukoje, line 197: It would be good to align the message with the timeout on line 203, or vice versa.

danijelvukoje (approved):
Great, works better 🙂

src/contexts/RepositoryEventsProvider.tsx
gularsson: Yeah, that was a mistake :slightly_frowning_face:

dannelundqvist (approved):
👍

PR #1251, opened 2025-03-31
fix: tweak scroll behaviour, wire with preview and table end or start…

gularsson:
… reached

### DIFF
--- a/src/components/Table/WireRow.tsx
+++ b/src/components/Table/WireRow.tsx
@@ -40,7 +40,7 @@ export const WireRow = ({ row, handleOpen, openDocuments, type }: {
     <TableRow
       tabIndex={0}
       className={cn(
-        'flex cursor-default snap-start scroll-mt-[70px] ring-inset focus:outline-none focus-visible:ring-2 focus-visible:ring-table-selected data-[state=selected]:bg-table-selected',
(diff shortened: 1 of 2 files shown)


src/components/Table/WireRow.tsx
```diff
@@ -40,7 +40,7 @@ export const WireRow = ({ row, handleOpen, openDocuments, type }: {
     <TableRow
       tabIndex={0}
       className={cn(
-        'flex cursor-default scroll-mt-10 ring-inset focus:outline-none focus-visible:ring-2 focus-visible:ring-table-selected data-[state=selected]:bg-table-selected',
+        'flex cursor-default snap-start scroll-mt-[70px] ring-inset focus:outline-none focus-visible:ring-2 focus-visible:ring-table-selected data-[state=selected]:bg-table-selected',
```
Victor-Lindh, line 43: I can't get the snapping behaviour to work, unfortunately.

gularsson: I'll remove it, the issue was that we didn't show the entire row (when a preview is open) after introducing the "search" header.

Victor-Lindh (approved):

dannelundqvist (approved):

PR #1254, opened 2025-03-31
cancel the previous effect (loop) when access token changes

danijelvukoje:


gularsson (approved):
Works like a charm :+1:

PR #1262, opened 2025-04-01
Duplicating event must not alter data

Victor-Lindh:


shared/transformations/lib/flattenYMap.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export function flattenYMap(ymap: Y.Map<unknown>): Result {
+  const result: Result = {}
+
+  for (const [key, value] of ymap.entries()) {
+    if (value instanceof Y.Map) {
+      result[key] = flattenYMap(value)
+    } else if (value instanceof Y.Array) {
+      result[key] = value.toArray()
+        .map((item) => item instanceof Y.Map ? flattenYMap(item) : item)
```
gularsson, line 16: To get rid of the lint warning:
```suggestion
        .map((item): unknown => item instanceof Y.Map ? flattenYMap(item) : item)
```

gularsson (approved):
:+1:
Maybe we could rename the function :thinking:

gularsson (approved):

PR #1266, opened 2025-04-02
Feature/ele 1229 multiple author documents are created

gularsson:


src/components/Init/lib/actions/author.ts
```diff
@@ -6,14 +6,26 @@ import { BoolQueryV1, QueryV1, TermQueryV1 } from '@ttab/elephant-api/index'
...
+ * @throws If the author document cannot be created or updated.
+ */
 export async function initializeAuthor({ url, session, repository }: {
   url: URL
   repository: Repository
   session: Session
 }): Promise<true> {
   try {
     const client = new Index(url.href)
+    const envRole = url.href.includes('stage') ? 'stage' : 'prod'
```
Victor-Lindh, line 28: Maybe this check should be by `.stage.` instead? Or some `meta.env` key, even?
You never know what we might add to the uri, all I'm saying... 😄

gularsson: afaik meta.env are set at build time with vite, so that's not an option I believe.

Victor-Lindh: I think they are, if they are prefixed with `VITE_`.
But you might as well check against the hostname instead.

Victor-Lindh (approved):

PR #1284, opened 2025-04-07
feat(editorial-info): add support for new deliverable type

gularsson:
- Introduced `editorial-info` as a new deliverable type.
- Added `editorialInfoDocumentTemplate` for generating templates.
- Updated workflow specifications for `editorial-info`.
- Enhanced `Assignment` and `AssignmentRow` components to handle `editorial-info`.
- Modified `CreateDeliverablePrompt` to support `editorial-info`.
- Renamed `getTemplate` to `getTemplateFromView` for clarity.
- Added `getDeliverableType` and `getTemplateFromDeliverable` utilities.

### DIFF
--- a/src/defaults/assignmentTypes.tsx
+++ b/src/defaults/assignmentTypes.tsx
@@ -29,7 +29,7 @@ export const AssignmentTypes: DefaultValueOption[] = [
     iconProps
   },
   {
-    label: 'Tillred',
+    label: 'Till red',
     value: 'editorial-info',
     icon: FileWarning,
     iconProps
--- a/src/defaults/assignmentTypes.tsx
+++ b/src/defaults/assignmentTypes.tsx
@@ -5,7 +5,8 @@ import {
   Camera,
   Video,
   Aperture,
-  ZapIcon
+  ZapIcon,
+  FileWarning
 } from '@ttab/elephant-ui/icons'
 
 const iconProps = {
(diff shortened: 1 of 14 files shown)


danijelvukoje (commented):
Looks good, I have noticed only one issue. Once created Tillred document doesn't get saved.

src/defaults/templates/editorialInfoDocumentTemplate.ts
```diff
@@ -0,0 +1,47 @@
...
+        }
+      }
+    ],
+    meta: [
+      {
+        type: 'core/note',
+        data: {
+          text: 'Obs! Detta meddelande är inte avsett för publicering.'
+        },
+        role: 'public'
```
Victor-Lindh, line 39: Should the role here be `internal`, perhaps?

src/defaults/assignmentTypes.tsx
```diff
@@ -27,6 +28,12 @@ export const AssignmentTypes: DefaultValueOption[] = [
     icon: ZapIcon,
     iconProps
   },
+  {
+    label: 'Tillred',
```
Victor-Lindh, line 32: Here, the term 'Tillred' is present, while in other places there are 'till red'.
One common label would be great!
(Personally, I prefer the variant used here 😄 )

src/defaults/templates/editorialInfoDocumentTemplate.ts
gularsson: It's public, in the way that it's "published" to the customers, however they in turn should not publish this to their readers

src/defaults/assignmentTypes.tsx
gularsson: I do prefer it as well, so I started using it, however i noticed that "till red" was used so i changed, but forgot one.

PR #1286, opened 2025-04-07
Feature open in readonly mode

Victor-Lindh:


danijelvukoje (commented):
It is not possible to scroll the published articles.
It's possible to remove the images from the article, it doesn't take any effect but it might be nice to disable that as well.

Victor-Lindh:
> 1. It is not possible to scroll the published articles.
> 2. It's possible to remove the images from the article, it doesn't take any effect but it might be nice to disable that as well.
1. Ok, I'll have a look.
2. That should be mitigated in the latest release of textbit-plugins, did you run `npm i` first?

danijelvukoje:
Right, after updating textbit-plugins, images are read only as well.

danijelvukoje (approved):
Nice, just noticed while testing that fact boxes are removable (with no effect), same like images were before updating textbit-plugins.

gularsson (commented):
I'm _only_ able to open in read-only mode now :thinking:

gularsson:
> > I'm _only_ able to open in read-only mode now 🤔
> 
> I was under the impression that published articles was to be opened in read-only by default. In case you'd want to go into edit mode, it has to be an active choice by the user (creating a new draft)?

Skill issue on my part

gularsson (approved):
:+1:

PR #1287, opened 2025-04-07
Feature/ele 1232 create article dialog should only show future plannings

gularsson:


Victor-Lindh (commented):
While testing, I noticed that I can't scroll vertically in the ComboBox.
I guess it's missing the `modal={asDialog)` prop

src/components/Move/index.tsx
```diff
@@ -190,6 +191,12 @@ export const Move = (props: ViewProps & {
                   </Button>
                 </>
               )}
+              <Checkbox
+                id='SearchOlder'
+                defaultChecked={searchOlder}
+                onCheckedChange={(checked: boolean) => { setSearchOlder(checked) }}
+              />
+              <Label htmlFor='SearchOlder' className='text-muted-foreground'>Även äldre</Label>
```
Victor-Lindh, line 199: Super picky, but I think `Visa äldre` sounds better here 😬
Feel free to ignore if you disagree.

src/lib/datetime.ts
```diff
@@ -188,6 +188,36 @@ export function dateInTimestampOrShortMonthDayTimestamp(date: string, locale: st
...
+
+  const inputDate = new Date(date)
+  const today = new Date()
+
+  const isToday = inputDate.getDate() === today.getDate()
+    && inputDate.getMonth() === today.getMonth()
+    && inputDate.getFullYear() === today.getFullYear()
+
+  if (isToday) {
+    return 'i dag'
```
Victor-Lindh, line 208: Gold star ⭐ for keeping in line with TT-språket here

Victor-Lindh (approved):
👍🏻

PR #1288, opened 2025-04-08
Refactored open documents tracking into a new Hocuspocus extension.

dannelundqvist:
Should now work correctly with many redis instances and correctly handle startups/crashes et al.

### DIFF
--- a/src-srv/utils/extensions/OpenDocuments.ts
+++ b/src-srv/utils/extensions/OpenDocuments.ts
@@ -1,0 +1,199 @@
+import * as Y from 'yjs'
+import type {
+  afterUnloadDocumentPayload,
+  connectedPayload,
+  Extension,
+  onDisconnectPayload
+} from '@hocuspocus/server'
+import logger from '../../lib/logger.js'
+import type { RedisCache } from '../RedisCache.js'
+
+interface EleContext {
+  user: {
+    sub: string
+    preferred_username: string
+    name: string
+  }
+}
+
+interface EleConnectedPayload extends connectedPayload {
+  context: EleContext
+}
+
+interface EleOnDisconnectPayload extends onDisconnectPayload {
+  context: EleContext
+}
+
+type CollaborationSnapshot = Record<string, {
+  id: string
+  users: Record<string, {
+    id: string
+    userName: string
+    name: string
+    count: number
+  }>
+}>
+
+/**
+ * OpenDocuments extension for Hocuspocus
+ *
+ * Tracks which documents are open and which users have them open.
+ * Maintains this data in a special 'document-tracker' Y.Doc.
+ */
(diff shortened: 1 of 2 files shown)


dannelundqvist:
> Neat! Did you mean that it should work correctly with multiple elephant-chrome instances?

Yes, the `onConfigure()` method first tries to fetch data from an already connected redis instance. It only creates a new tracking document and put in redis if there is none before (when starting the system). The tracking document is then allowed to be stored in redis and synced to other instances. I'd love to so it tested (I haven't).

danijelvukoje:
Okay, and how do we handle a scenario where an instance crashes? Do we clean up the document-tracker document?

dannelundqvist:
> Okay, and how do we handle a scenario where an instance crashes? Do we clean up the document-tracker document?

Very good question. It does not handle that case. I'll test it more...

dannelundqvist:
> Okay, and how do we handle a scenario where an instance crashes? Do we clean up the document-tracker document?

This should be be addressed with the latest additions tracking server instances and removes all references if a server instance is lost.

src-srv/utils/extensions/OpenDocuments.ts
```diff
@@ -158,19 +260,25 @@ export class OpenDocuments implements Extension {
...
-        yUser.set('count', count - 1)
+        // Find first occurrence of current instance
+        const n = count.toJSON().findIndex((v) => v === this.#instanceId)
+        if (n !== -1) {
+          count.delete(n)
+        } else {
+          logger.warn(`Client <${userId}> on instance <${this.#instanceId}> disconnected from <${documentName}> but there was no instance entry in count`)
+        }
       }
       yDocEntry.set('count', yDocEntry.get('count') ?? 0 - 1)
```
danijelvukoje, line 280: This should be
```yDocEntry.set('count', (yDocEntry.get('count') ?? 0) - 1)```

File: src-srv/utils/extensions/OpenDocuments.ts [hunk 2]
```diff
@@ -122,8 +218,11 @@ export class OpenDocuments implements Extension {
...
-          // User already connected to document, increase connection count
-          yUser.set('count', yUser.get('count') ?? 0 + 1)
+          // User already connected to document, increase connection count,
+          // by adding the instance id to the count array. Each instance can
+          // appear multiple times and increase the count accordingly.
+          const count = yUser.get('count') as Y.Array<string>
+          count.push([this.#instanceId])
         }
 
         // Increase total document connection count
```
danijelvukoje, line 225: This should be
yDocEntry.set('count', (yDocEntry.get('count') ?? 0) + 1)

danijelvukoje (approved):
Looks great now! Very good idea to move and handle this logic in an extension.

PR #1296, opened 2025-04-10
Fix: remove fullday column in Approvals view, text/flash/editorial-info assignments should not allow full day publish time, update publish time on status change if time has passed

Victor-Lindh:


dannelundqvist:
Should it be possible to set the the text assignment to full day?

Skärmavbild 2025-04-10 kl  16 23 11

danijelvukoje:
Seems to be possible for both article and flash assignments.

Victor-Lindh:
> Seems to be possible for both article and flash assignments.

Made new commit, should work now

danijelvukoje (approved):
👍

dannelundqvist (approved):
Nice!

PR #1303, opened 2025-04-11
fix(AssignmentRow): improve dropdown menu and assignment row UX

gularsson:
- Ensured `onPrimary` in AssignmentRow prevents default and stops propagation.


Victor-Lindh (approved):

Victor-Lindh (commented):
Found that clicking outside the prompt also fires the onPrimary handler

PR #1305, opened 2025-04-14
Feature/ele 1250 approvals view should display users having docs open

dannelundqvist:
Also fixed the syncing of open documents between server instances, as it turned out, did not work.

### DIFF
--- a/src-srv/utils/extensions/OpenDocuments.ts
+++ b/src-srv/utils/extensions/OpenDocuments.ts
@@ -2,11 +2,13 @@ import * as Y from 'yjs'
 import type {
   afterUnloadDocumentPayload,
   connectedPayload,
+  DirectConnection,
   Extension,
+  Hocuspocus,
+  onConfigurePayload,
   onDisconnectPayload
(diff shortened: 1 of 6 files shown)


src-srv/utils/extensions/OpenDocuments.ts
```diff
@@ -286,24 +260,39 @@ export class OpenDocuments implements Extension {
...
 
   /**
-   * Get a snapshot of all tracked documents and their users.
+   * Get reference to the Yjs document
    */
-  getSnapshot(): CollaborationSnapshot {
-    const yOpenDocuments = this.#doc.getMap('open-documents')
-    const yInstances = this.#doc.getMap('instances')
+  getDocument(): Y.Doc {
+    return this.#doc
```
danijelvukoje, line 275: How will this work if `this.#doc` is never reassigned after it's initialised to an empty Y doc?

danijelvukoje (approved):
Looks great. Thanks for the briefing on how collaboration works in the document tracker.

gularsson (approved):

PR #1312, opened 2025-04-15
feat: add Baboon print article creation support

gularsson:
- Introduced Baboon backend integration with new environment variable `BABOON_URL`.
- Added `Baboon` class to handle print article creation via Twirp API.
- Updated `RegistryProvider` to include Baboon client initialization.
- Enhanced `AssignmentRow` with a new menu option to create print articles.
- Implemented `CreatePrintArticle` component for user interaction.
- Added `useFetchPrintFlows` hook to fetch print flows using SWR.
- Updated `DatePicker` to support a disabled state.
- Modified server URL fetching logic to include `BABOON_URL`.

This change enables users to create print articles directly from the application.


gularsson:
Copilot was unintentional ;)

sinusbrus (approved):
As far as i can tell the code looks good. I've testet the function locally and it works fine.

I will merge some code from this PR before I'm finished with mine.

sinusbrus:
I approved but see now that the PR has conflicts to be resolved.

PR #1325, opened 2025-04-22
Search view improvements

Victor-Lindh:


src/components/Datepicker.tsx
```diff
@@ -85,7 +87,7 @@ export const DatePicker = ({ date, changeDate, setDate, forceYear = false }: {
             if (!selectedDate) return
 
             if (changeDate) {
-              changeDate(undefined, { from: format(selectedDate, 'yyyy-MM-dd') }, 'self')
+              changeDate(undefined, { ...(keepQuery && { ...keepQuery }), from: format(selectedDate, 'yyyy-MM-dd') }, 'self')
```
gularsson, line 90: keepQuery gets confusing, when passed to `DateChanger` it's a `boolean` when passed to `DatePicker` it's a `QueryParam`.

Could we use `DatePicker` directly and circumvent the `Search` specific additions?

On another note, I'm wondering whether `useLink` is the correct way? Do we want to add a navigation? Maybe when navigating in `Plannings` but not here? But I can't see it making a navigation either :thinking: 

useQuery/setQuery, might be better. But that's another discussion.

src/views/SearchOverview/SearchColumns.tsx
```diff
@@ -1,177 +1,125 @@
...
-      cell: ({ row }) => {
-        const status = row.getValue<string>('documentStatus')
-        const type = row.getValue<string>('itemType')
+  authors?: IDBAuthor[]
+  organisers?: IDBOrganiser[]
+  type: 'plannings' | 'events' | 'assignments' | 'articles'
+}): SearchColumnType {
+  if (type === 'plannings') {
+    return [...planningListColumns({ sections, authors })]
+  }
```
gularsson, line 32: We should probably make sure that _all_ search results have a clear date shown, now not all views have that Events for example.

```
return [dateColumn, ...planningListColumns(...)]
```

src/views/SearchOverview/search.ts
```diff
@@ -47,44 +50,48 @@ export const search = ({
...
   accessToken,
   indexUrl,
   setData,
-  status
+  status,
+  query
 }: Props): void => {
   void (async () => {
-    if (!accessToken || status !== 'authenticated' || (!text || text?.length < 1)) {
+    if (!accessToken || status !== 'authenticated'/*  || (!text || text?.length < 1) */) {
```
gularsson, line 61: remove?

gularsson (changes requested):

src/components/Datepicker.tsx
Victor-Lindh: Made new commit, got rid of the Search-specifics, as well as the useLink that wasn't needed.

Victor-Lindh: Made new commit that should fix this.

src/views/SearchOverview/SearchColumns.tsx
Victor-Lindh: Added new commit for this

gularsson (approved):

src/components/Datepicker.tsx
```diff
@@ -87,7 +89,7 @@ export const DatePicker = ({ date, changeDate, setDate, forceYear = false, disab
...
```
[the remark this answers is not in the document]
Victor-Lindh: No, I simply forgot to remove this part. Fixed in new commit.

PR #1326, opened 2025-04-22
fix(flash): add publish timestamp to flash creation

gularsson:
Added a `publish` field to the flash creation payload, which stores the current timestamp in ISO format. This ensures that the flash will show up in Approvals view


danijelvukoje (approved):
✅

PR #1331, opened 2025-04-23
fix(Approvals): remove duplicate active-users-avatars

gularsson:


Victor-Lindh (commented):
Now I can't see these avatars either (furthest to the right in this image), is that intentional?

Skärmavbild 2025-04-24 kl  14 08 53

Victor-Lindh (approved):
👍🏻

PR #1333, opened 2025-04-24
fix: factbox should be removable in Editor

Victor-Lindh:


gularsson (approved):
However it seems like the remove X is not showing

PR #1334, opened 2025-04-24
fix: correctly save image caption and byline

Victor-Lindh:


shared/transformations/newsdoc/tt/visual.ts
```diff
@@ -39,12 +39,20 @@ export const transformVisual = (element: Block): TBElement => {
...
-  const { id, properties } = element
+  const { id, properties, children } = element
+  const textNode = children?.find((c) => c.type === 'tt/visual/text')
+  let text = ''
+  if (textNode && 'children' in textNode && textNode?.children && Array.isArray(textNode.children)) {
+    const [child] = textNode?.children ?? { text: '' }
+    if ('text' in child) {
+      text = child?.text
+    }
+  }
```
danijelvukoje: Slight redundancy, something like this should suffice and it's easier to read:
```suggestion
  if (textNode && Array.isArray(textNode.children)) {
    const [child] = textNode.children
    if (child && 'text' in child) {
      text = child.text ?? ''
    }
  }
```

Victor-Lindh: Added new commit with these suggestions

shared/transformations/newsdoc/tt/visual.ts
```diff
@@ -39,17 +40,35 @@ export const transformVisual = (element: Block): TBElement => {
...
+  const textNode = children?.find((c) => c.type === 'tt/visual/text')
+  const bylineNode = children?.find((c) => c.type === 'tt/visual/byline')
+
+  function getText(node: Descendant | undefined) {
+    let text = ''
+    if (node && 'children' in node && Array.isArray(node?.children)) {
+      const [child] = node.children
+
+      if (child && 'text' in child) {
+        text = child?.text
```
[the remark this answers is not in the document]
Victor-Lindh: Since we are initializing text to an empty string above the if-statement, I prefer to not reassign it to an empty string again, should the text not be defined.
For stylistic purposes, I'd prefer to add a check to the if clause, then remove the op-chain. Seems cleaner.

gularsson (approved):
:+1:

PR #1339, opened 2025-04-25
Version info improvements

Victor-Lindh:


src/components/Version/index.tsx
```diff
@@ -54,6 +57,45 @@ export const Version = ({ documentId, hideDetails = false }: { documentId: strin
       return statuskeys.some((key) => STATUS_KEYS.includes(key))
     })
 
+    const fetchDoc = async (v: DocumentVersion) => {
+      // Used to fetch the previous document version in order to get hold of the title,
+      // that can be displayed in the list of previous versions.
+      const response = await fetch(`${BASE_URL}/api/documents/${documentId}?version=${v.version}`)
```
danijelvukoje, line 63: Should we use the repository instance here instead of fetching ad hoc?

PR #1344, opened 2025-04-28
set parent event title as title

Victor-Lindh:


src/views/Event/components/PlanningTable.tsx
```diff
@@ -91,7 +91,7 @@ export const PlanningTable = ({ provider, documentId, asDialog }: {
                   'core/event': [Block.create({
                     type: 'core/event',
                     uuid: documentId,
-                    title: 'Untitled',
+                    title: payload?.title || 'Untitled',
```
dannelundqvist, line 94: As everything is in Swedish maybe it would be better to name it something like 'Ny händelse'?

danijelvukoje (approved):

danijelvukoje (approved):
Great, that's more semantically correct.

dannelundqvist (approved):

PR #1356, opened 2025-04-30
Approvals view cards should display title from article, also set document.title to heading-1 value

Victor-Lindh:


gularsson (approved):
:+1:

PR #1358, opened 2025-04-30
Feature/ele 1310 scheduled articles should appear in approvals view

dannelundqvist:
* Correctly fetch text assignment texts based on publish time in Approvals view.
* Fix faulty utc from/to span.
* Added functions that should deprecate other time/date functions including using new Date() which won't give you the local datetime in the systems configured timeZone if using the system in another timezone.

### DIFF
--- a/src/hooks/index/lib/assignments/fetchAssignments.ts
+++ b/src/hooks/index/lib/assignments/fetchAssignments.ts
@@ -183,6 +183,7 @@ function constructQuery(
   timeZone: string
 ): QueryV1 {
   const { from, to } = getUTCDateRange(date, timeZone)
+  // CAVEAT: This is in reality a local date but with zero time and a faulty Z in it.
   const formattedDate = format(date, 'yyyy-MM-dd\'T00:00:000Z\'')
 
   if (dateType === 'publish') {
--- a/src/hooks/index/lib/assignments/fetchAssignments.ts
+++ b/src/hooks/index/lib/assignments/fetchAssignments.ts
@@ -183,8 +183,8 @@ function constructQuery(
   timeZone: string
 ): QueryV1 {
   const { from, to } = getUTCDateRange(date, timeZone)
-  // CAVEAT: This is in reality a local date but with zero time and a faulty Z in it.
-  const formattedDate = format(date, 'yyyy-MM-dd\'T00:00:000Z\'')
+  // CAVEAT: This is in reality a local date but with zero time and a misleading Z in it.
+  const formattedDate = format(date, 'yyyy-MM-dd\'T00:00:00:000Z\'')
 
   if (dateType === 'publish') {
     return QueryV1.create({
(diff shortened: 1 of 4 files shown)


src/views/Approvals/index.tsx
```diff
@@ -45,18 +47,18 @@ export const ApprovalsView = (): JSX.Element => {
...
-    getDateTimeBoundariesUTC(typeof query.from === 'string'
-      ? new Date(`${query.from}T00:00:00.000Z`)
-      : new Date()
-    ), [query.from])
-
+  const date = useMemo(() => {
+    return (typeof query.from === 'string')
+      ? newLocalDate(timeZone, { date: query.from }) as Date
+      : newLocalDate(timeZone) as Date
+  }, [query.from, timeZone])
```
gularsson, line 54: Seems wrong to have to cast it as Date?
It could be overloaded so return type would be inferred from options passed. Something like this maybe (undefined is superfluous in the switch?):
```
export function newLocalDate(
  timeZone: string,
  options: { as: 'datetime', date?: string }
): string
export function newLocalDate(
  timeZone: string,
  options: { as: 'date', date?: string }
): Date
export function newLocalDate(
  timeZone: string,
  options?: { as?: 'datetime' | 'date', date?: string }
): Date
export function newLocalDate(
  timeZone: string,
  options?: { as?: 'datetime' | 'date', date?: string }
): string | Date {
  const inputDate = options?.date
    ? fromZonedTime(`${options.date}T00:00:00`, timeZone)
    : new Date()

  const date = toZonedTime(inputDate, timeZone)

  switch (options?.as) {
    case 'datetime':
      return format(date, 'yyyy-MM-dd\'T\'HH:mm:ss')

    case 'date':
    default:
      return date
  }
}

```

src/hooks/index/lib/assignments/fetchAssignments.ts
```diff
@@ -169,22 +174,30 @@ export async function fetchAssignments({ index, repository, type, requireDeliver
...
-
-  const formattedDate = `${year}-${month}-${day}T${hour}:${minute}:${second}Z`
+function constructQuery(
+  date: Date,
+  dateType: 'start-date' | 'publish',
+  timeZone: string
+): QueryV1 {
+  const { from, to } = getUTCDateRange(date, timeZone)
+  // CAVEAT: This is in reality a local date but with zero time and a faulty Z in it.
+  const formattedDate = format(date, 'yyyy-MM-dd\'T00:00:000Z\'')
```
gularsson, line 187: Wouldn't this search be some hours of depending on timediff to utc? :thinking:

gularsson (changes requested):

src/hooks/index/lib/assignments/fetchAssignments.ts
gularsson: Any how, isn't `'yyyy-MM-dd\'T00:00:00.000Z\'') correct zero time string?

gularsson: ah I see... i think...
```          "document.meta.core_assignment.data.start": [
            "2025-05-01T22:00:00Z"
          ],
          "document.meta.core_assignment.data.start_date": [
            "2025-05-02T00:00:00Z"
          ],
``

gularsson (approved):

Victor-Lindh (approved):

PR #1362, opened 2025-05-02
fix: create new dialog instead of overwriting current

Victor-Lindh:
Improvement from before, when flow of creating multiple documents in a row (event -> planning -> assignment) wouldn't work because of planning modal effectively removing event modal, thus never allowing event to be created at all


dannelundqvist (approved):
👍

PR #1366, opened 2025-05-05
fix(WorkflowStatus): do snapshot before setting status

gularsson:
- Introduced a `force` query parameter to the snapshot API to allow forced snapshots even when no changes are detected.
- Updated `snapshotDocument` method in `CollaborationServer` to handle `force` parameter and ensure proper hash usage.
- Enhanced `useWorkflowStatus` hook to utilize the new snapshot response for version updates.
- Refactored `snapshot` utility to support the `force` parameter and return detailed response data.
- Adjusted `Editor` view to correctly handle version logic based on workflow status.

These changes improve snapshot functionality and ensure accurate versioning in workflows.
```


dannelundqvist (approved):
👍

PR #1371, opened 2025-05-06
fix(wires): search in content

gularsson:


danijelvukoje (approved):
Should we also search in other content types? Otherwise 👍

Victor-Lindh (approved):

danijelvukoje (approved):
Excellent!

PR #1372, opened 2025-05-06
fix: remove draft status transition in events, plannings

Victor-Lindh:


gularsson (approved):
:+1: 

Screenshot_2025-05-06-15-03-02_17020
Could perhaps remove the with bar below (if easy, if time)

Victor-Lindh:
> Could perhaps remove the white bar below (if easy, if time)

It's there because of the `cancelled` status, which is not in the specification yet.
Separate issue.

PR #1375, opened 2025-05-06
fix(useAssignments): extend event handling and logging

gularsson:
- Extended `useRepositoryEvents` to include `core/article` and `core/article+meta` event types.
- Updated assignment matching logic to consider `_deliverableId` for event UUIDs.

This improves event handling.


Victor-Lindh (approved):
✅

PR #1377, opened 2025-05-07
fix(factbox): add body wrapper for factbox content

gularsson:
- Updated `transformFactbox` to wrap content in a `core/factbox/body` block with a unique ID and class.
- Modified `revertFactbox` to extract and process the `core/factbox/body` block instead of all children except the title.


Victor-Lindh (approved):
👍🏻

PR #1379, opened 2025-05-07
fix(latest): enhance document handling and UI

gularsson:
Replace `useAssignments` with `useDocuments` for fetching data in `Latest` view.


Victor-Lindh (approved):
👍🏻

PR #1383, opened 2025-05-07
fix(editor): add button to open latest version from read-only

gularsson:
When in readOnly mode and displayed version isn't the latest, instead of StatusMenu show button to go-to-latest.

Problem apparent with: https://github.com/ttab/elephant-chrome/pull/1379

### DIFF
--- a/src/views/Editor/EditorHeader.tsx
+++ b/src/views/Editor/EditorHeader.tsx
@@ -1,4 +1,4 @@
-import { useHistory, useNavigation, useView, useWorkflowStatus, useYValue } from '@/hooks'
+import { useHistory, useLink, useNavigation, useView, useWorkflowStatus, useYValue } from '@/hooks'
 import { Newsvalue } from '@/components/Newsvalue'
 import { useCallback, useEffect, useRef, useState } from 'react'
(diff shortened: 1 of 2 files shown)


src/views/Editor/EditorHeader.tsx
```diff
@@ -103,6 +106,9 @@ export const EditorHeader = ({ documentId, readOnly, readOnlyVersion }: { docume
   }, [deliverablePlanning, dispatch, documentId, history, state.viewRegistry, viewId, workflowStatus])
 
   const title = documentType === 'core/editorial-info' ? 'Till red' : 'Artikel'
+
+  const isReadOnlyAndUpdated = workflowStatus?.name !== 'usable' && readOnly
```
Victor-Lindh, line 110: For a split second, the "Gå till senaste versionen" button is visible when the Editor is loading, due to the `workflowStatus` being `undefined`.
An extra check seems to do the trick:
```suggestion
  const isReadOnlyAndUpdated = workflowStatus && workflowStatus?.name !== 'usable' && readOnly
```

Victor-Lindh (changes requested):

Victor-Lindh (approved):
👍🏻

PR #1384, opened 2025-05-07
feat(editorial-info): add support for editorial info types

gularsson:
Introduced a new `EditorialInfoTypes` component to handle the selection of editorial info types. Integrated this component into the `MetaSheet` view, conditionally rendering it when the document type is `core/editorial-info`.

Changes include:
- Created `EditorialInfoTypes` in `src/components`.
- Updated `MetaSheet` to include the new component with proper labeling.

This enhances the editor's functionality by allowing users to manage editorial info types directly within the UI.

Edit:
Added the `EditorialInfoType` to IndexedDB and cache, wondering about the migration though... doesn't seem to work with out deleting the db.


danijelvukoje (commented):
Works well, it would be nice if a user could unselect the info type and the bottom of the meta sheet is a bit out of the viewport.

gularsson:
> Works well, it would be nice if a user could unselect the info type and the bottom of the meta sheet is a bit out of the viewport.

Fixed

danijelvukoje (approved):

PR #1386, opened 2025-05-08
feat(assignment): add visibility control for assignments

gularsson:
- Updated `AssignmentType` to handle visibility for 'flash' assignments.
- Introduced `AssignmentVisibility` component for managing visibility.
- Adjusted `assignmentPlanningTemplate` to set visibility based on type.
- Ensured `createFlash` sets flashes as private by default.
- Integrated `AssignmentVisibility` into `Assignment` and `AssignmentRow`.
- Refactored paths and logic to support new visibility feature.

This change improves assignment management by adding visibility control.


Victor-Lindh (approved):
Nice, I think it works well!
Only thing I'm not entirely sure about are the icons for public vs internal assignments,
as they are the same as the document status icons, might be confusing.
But I don't really have any better suggestions.

Victor-Lindh (commented):
Skärmavbild 2025-05-09 kl  10 06 33
One small bug, is that when creating a flash assignment in a planning, the 'internal' is preselected, but in the Dropdown the public one is instead.

gularsson:
> Nice, I think it works well! Only thing I'm not entirely sure about are the icons for public vs internal assignments, as they are the same as the document status icons, might be confusing. But I don't really have any better suggestions.

Yeah, that was my concern as well, but had no other option other than introducing yet another icon.

PR #1394, opened 2025-05-09
Properly parse meta heads statuses and versions

dannelundqvist:
_I know the tests might need improvement... but there's a few tests at least._


__tests__/getStatusFromMeta.test.ts
```diff
@@ -0,0 +1,164 @@
...
+      name: 'draft',
+      version: BigInt(0),
+      creator: 'core://user/abc',
+      cause: undefined
+    })
+  })
+
+  it('returns cancelled when version is -1', () => {
+    const meta = {
+      currentVersion: BigInt(-1),
```
hugowetterberg, line 23: Current version of a document can never be less that 1.

hugowetterberg (changes requested):

PR #1396, opened 2025-05-11
fix(planning/event): sync end date with start date on change

gularsson:
Update PlanDate component to set both start and end date fields when the date is changed. This ensures that the end date is always synchronized with the start date, improving consistency in planning item data. Also, use the user's time zone when creating new dates for better accuracy.


Victor-Lindh (approved):
✅

PR #1405, opened 2025-05-14
Remove excessive empty paragraphs on snapshotting text documents

dannelundqvist:
Removes consecutive empty paragraphs for text documents (article, flash and editorial-info). Do not touch anything but paragraphs (role is empty).


gularsson (commented):
This only applies the removal on save, which I guess is fine, however it would probably be more transparent to the user to have it done on the collaborative document, before save? :thinking:

gularsson (approved):

PR #1411, opened 2025-05-15
Feature/ele 1354 pollsubscribe impl should ignore events not in the result

gularsson:


src/hooks/index/useDocuments/index.ts
```diff
@@ -122,7 +123,10 @@ export const useDocuments = <T extends HitV1, F>({ documentType, query, size, pa
           }
 
           console.error('Polling error:', error)
-          toast.error('Polling failed. Retrying...')
+          toast.error('Misslyckades att automatiskt uppdatera listan. Försöker igen om 30 sekunder.')
+
+          // Wait before next retry if failed
+          await new Promise((resolve) => setTimeout(resolve, 30000))
```
dannelundqvist, line 129: It's good to avoid retry timeouts that potentially can hammer a backend from all clients at the same time when recovering from issues. There is a (naive) jitter function In `src-srv/utils/extensions/OpenDocuments.ts`. Maybe we could add that as a shared function and call it here as well to at least avoid the some potential issues?

```
  #getInterval(min: number, max: number) {
    const minCeiled = Math.ceil(min)
    const maxFloored = Math.floor(max)
    return Math.floor(Math.random() * (maxFloored - minCeiled + 1) + minCeiled) * 1000
  }
```

dannelundqvist (changes requested):
A minor comment on the retry timeout only.

src/hooks/index/useDocuments/index.ts
gularsson: Yea, probably, didn't want to get to clever, so I used a long wait. My hopes are that this should only cause one request per client every 30 sec, so it shouldn't be that many.
I'll look into jittering it a bit.

dannelundqvist (approved):

PR #1419, opened 2025-05-16
Remove cancelled status, replace with unpublished, improve labels

Victor-Lindh:


gularsson (approved):
I'm seeing some errors from Twirp in events, that is hopefully fixed in main. :crossed_fingers: 
Also seeing first publish of a planning requiring a cause, i think that is solved in my factbox PR. :crossed_fingers: 

Looks good other than that.

PR #1420, opened 2025-05-16
feat(factbox): track and propagate document change state

gularsson:
- Add `isChanged` prop to `FactboxHeader` and propagate from `FactboxContainer` using `useYValue` to track document changes.
- Update `StatusMenu` to use `isChanged` for determining cause and requirement logic.
- Pass `onChange` handlers to `FactboxContent` and `Textbit.Editable` to update change state when content is modified.
- Set `debounce={0}` on `Textbit.Root` - only way to detect change.


Victor-Lindh (approved):
Works well, from what I can tell 👍🏻

PR #1421, opened 2025-05-16
feat: ignore unknown fields in Twirp clients

gularsson:
Add `jsonOptions: { ignoreUnknownFields: true }` to all TwirpFetchTransport client initializations in shared modules. This ensures that clients will gracefully handle extra fields in JSON responses, improving forward compatibility and robustness against backend changes.


Victor-Lindh (approved):
👍🏻

PR #1422, opened 2025-05-17
fix(wires): unique slugline validation'

gularsson:
A bit construed solution, but this will cause irrevocable damage to the planning, since the slug can not be changed afterwards

### DIFF
--- a/src/views/Wire/WireViewContent.tsx
+++ b/src/views/Wire/WireViewContent.tsx
@@ -12,7 +12,8 @@ import { Button, Checkbox, ComboBox, Input, Label } from '@ttab/elephant-ui'
 import { CircleXIcon, Tags, GanttChartSquare, Cable, BriefcaseBusiness, Tag } from '@ttab/elephant-ui/icons'
 import { useCollaboration, useRegistry, useYValue } from '@/hooks'
 import { useSession } from 'next-auth/react'
-import { useRef, useState } from 'react'
+import type { PropsWithChildren } from 'react'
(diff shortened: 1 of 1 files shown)


src/views/Wire/WireViewContent.tsx
```diff
@@ -60,6 +62,7 @@ export const WireViewContent = (props: ViewProps & {
 
       <View.Content>
         <Form.Root asDialog={props.asDialog}>
+          {!!selectedPlanning && <ValidateNow />}
```
dannelundqvist, line 65: <ValidateNow> is a local component used once. It always returns null, and it never receives any props. The effect needs a prop to fire. I'm not sure I understand this code... it looks like it does nothing.

Victor-Lindh: I believe it receives props as a child of the `Form.Root` component.

Victor-Lindh (commented):
It does the job correctly for when selecting an already existing planning 👍🏻
However, manually adding an already existing slugline doesn't get validated.
Maybe that's a broader, separate feature than with creating articles from wires.

src/views/Wire/WireViewContent.tsx
gularsson: Yes, a bit construed as the comment says, but thats the only way i found to access the Form wrapper props outside of the wrapper

dannelundqvist: 🫣

Victor-Lindh (approved):
👍🏻

PR #1427, opened 2025-05-17
fix(snapshot): improve context validation and error handling

gularsson:
Enhance the Snapshot extension by introducing stricter context validation using `assertContext` and UUID checks for document names. Refactor `onDisconnect` to provide better error handling and clearer error messages. This improves robustness and prevents invalid operations during document storage and disconnection events.

### DIFF
--- a/src-srv/utils/extensions/Snapshot.ts
+++ b/src-srv/utils/extensions/Snapshot.ts
@@ -6,6 +6,7 @@ import {
 import { type DebouncedFunc, debounce } from 'lodash-es'
 import { assertContext } from '../../lib/assertContext.js'
 import { isValidUUID } from '../isValidUUID.js'
+import logger from '../../lib/logger.js'
 
 interface DefaultConfiguration {
   debounce: number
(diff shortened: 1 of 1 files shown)


src-srv/utils/extensions/Snapshot.ts
```diff
@@ -46,12 +56,16 @@ export class Snapshot implements Extension {
...
 
-    // If the document has a debounceFn it's dirty, check if there are no other clients
-    if (debouncedFn && payload.clientsCount === 0) {
+      // If the document has a debounceFn it's dirty, check if there are no other clients
+      if (debouncedFn && payload.clientsCount === 0) {
       // Flush (immidiately call the function)
-      await debouncedFn.flush()
+        await debouncedFn.flush()
+      }
+    } catch (ex) {
```
dannelundqvist, line 67: Catching an exception to throw it again hides the original stacktrace. Not sure this is a good idea.

dannelundqvist (changes requested):
Good thing checking the uuid validness. Questioning the value in catching error only to throw it again.

src-srv/utils/extensions/Snapshot.ts
gularsson: Yeah, I wanted to add context... but just handle and logging it should be enough

dannelundqvist (approved):

PR #1428, opened 2025-05-17
Feature/ele 1380 gracefully unload the pollsubscription using abortcontroller

gularsson:
Base is https://github.com/ttab/elephant-chrome/pull/1411~


src/hooks/index/useDocuments/index.ts
```diff
@@ -161,6 +161,12 @@ export const useDocuments = <T extends HitV1, F>({ documentType, query, size, pa
 
     const abortController = new AbortController()
 
+    const handleBeforeUnload = () => {
+      abortController.abort()
+    }
+
+    window.addEventListener('beforeunload', handleBeforeUnload)
```
dannelundqvist, line 168: As we are touting mobile (small screen) support it might be worth using `unload` event for safari on iOS as it is (unexplainable) still not supporting the `beforeunload` event. The unload event is deprecated and should not be used in other browser though.
https://bugs.webkit.org/show_bug.cgi?id=219102

dannelundqvist (changes requested):

src/hooks/index/useDocuments/index.ts
gularsson: :exploding_head:

dannelundqvist (approved):

PR #1429, opened 2025-05-18
feat(flash): add document status and tertiary action

gularsson:
- Introduced `CreateFlashDocumentStatus` type to handle different flash document statuses: 'usable', 'done', or undefined.
- Refactored `FlashViewContent` to use a prompt config array for handling send, save, and done prompts, reducing code duplication.
- Added tertiary button/action for "Godkänn flash" (approve flash).
- Updated `createFlash` to show different toast messages based on document status and improved error handling with user feedback.
- Updated form submit logic to require explicit button roles and handle tertiary actions.
- Updated newsvalue for created planning to 5

### DIFF
--- a/__tests__/createFlash.test.ts
+++ b/__tests__/createFlash.test.ts
@@ -56,7 +56,8 @@ describe('createFlash', () => {
       session,
       planning: { document: planning[1], id: planning[0] },
       hasSelectedPlanning: isExistingPlanning,
-      timeZone: 'Europe/Stockholm'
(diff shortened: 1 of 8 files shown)


dannelundqvist (changes requested):
Similar objections as for the event and planning. I would like to see
* "Skapa utkast"
* "Godkänn"
* "Publicera direkt!*

We don't need to have "Flash" in the buttons as the dialog is all about flash.

dannelundqvist (approved):

PR #1430, opened 2025-05-18
feat(latest): add flash and editorial-info types to view

gularsson:
Refactor Latest view to fetch and merge articles, flashes, and editorial info from the last 24 hours. Add logic to display "TT-FLASH" as the slugline for flash documents based on their URI.

### DIFF
--- a/src/views/Latest/hooks/useLatest.tsx
+++ b/src/views/Latest/hooks/useLatest.tsx
@@ -1,0 +1,142 @@
+import { useDocuments } from '@/hooks/index/useDocuments'
+import { useEffect, useMemo, useState } from 'react'
+import type { HitV1 } from '@ttab/elephant-api/index'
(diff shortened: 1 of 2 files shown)


dannelundqvist (approved):
This file is getting a bit out of hand. It would be more maintainable if the view itself knew nothing about how to call the repository. But It was already there and we are short on time and we know we probably will rework this in the future :). I think at least we should add a big FIXME to this file before merging.

gularsson:
> This file is getting a bit out of hand. It would be more maintainable if the view itself knew nothing about how to call the repository. But It was already there and we are short on time and we know we probably will rework this in the future :). I think at least we should add a big FIXME to this file before merging.

Moved the data fetching and massaging to a hook.

dannelundqvist (approved):

PR #1431, opened 2025-05-18
feat(event): add status handling to event submission

gularsson:
Enhance the EventViewContent form to support document status handling ('usable', 'done', or undefined) on submit. Updated the submit handler to accept a status parameter and send it with the stateless message. Modified the form footer to provide buttons for different submission statuses, improving user control over event publishing workflow.

Based on: https://github.com/ttab/elephant-chrome/pull/1429

### DIFF
--- a/src/views/Event/index.tsx
+++ b/src/views/Event/index.tsx
@@ -92,7 +92,9 @@ const EventViewContent = (props: ViewProps & { documentId: string }): JSX.Elemen
     // eslint-disable-next-line
   }, [provider])
 
-  const handleSubmit = (): void => {
+  const handleSubmit = ({ documentStatus }: {
+    documentStatus: 'usable' | 'done' | undefined
+  }): void => {
     if (props?.onDialogClose) {
       props.onDialogClose()
     }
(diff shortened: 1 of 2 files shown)


src/views/Event/index.tsx
```diff
@@ -163,11 +166,22 @@ const EventViewContent = (props: ViewProps & { documentId: string }): JSX.Elemen
...
-              <div className='flex justify-end px-6 py-4'>
+            <Form.Submit
+              onSubmit={() => handleSubmit({ documentStatus: 'usable' })}
+              onSecondarySubmit={() => handleSubmit({ documentStatus: 'done' })}
+              onTertiarySubmit={() => handleSubmit({ documentStatus: undefined })}
+            >
+              <div className='flex justify-between px-6 py-4'>
+                <div className='flex gap-2'>
+                  <Button type='button' variant='secondary' role='tertiary'>
+                    Spara
```
dannelundqvist, line 177: Looking at the naming of the other buttons I think this should be named "Skapa utkast" or "Utkast" to avoid confusion. Because regardless of which button you press it is saved.

dannelundqvist (changes requested):
Works good! I think we should avoid the word "Spara" to avoid confusion.

dannelundqvist (approved):

PR #1443, opened 2025-05-19
feat(modal): improve dialog close UX and escape handling

gularsson:
- Added confirmation prompt when closing dialogs with unsaved changes, both via close button and Escape key.
- Removed global Escape key handler from ModalProvider to prevent unintended modal closures.
- Updated ModalDialog to prevent default Escape key behavior.
- Improved Prompt dialog stacking with z-index.
- Minor style adjustment in Planning view dialog actions.

### DIFF
--- a/src/components/Move/index.tsx
+++ b/src/components/Move/index.tsx
@@ -147,7 +147,7 @@ export const Move = (props: ViewProps & {
           )}
         </ViewHeader.Content>
 
-        <ViewHeader.Action onDialogClose={props.onDialogClose}>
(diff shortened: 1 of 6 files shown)


Victor-Lindh (commented):
Works really well!
Minor thought of improvement is to be able to use the keyboard to select the primary button in the confirmation prompt.

Victor-Lindh (approved):

dannelundqvist (approved):
Works well!

(_There's an issue with using escape in the create planning dialog opened from within the create event dialog. But this is not due to this change, I have documented it in a new issue – ELE-1393_)

PR #1448, opened 2025-05-20
Babian middle release2

sinusbrus:


src/views/PrintArticles/PrintArticlesList.tsx
```diff
@@ -26,11 +28,21 @@ import type { ColumnDef } from '@tanstack/react-table'
 export const PrintArticleList = ({ columns }: {
   columns: ColumnDef<PrintArticle, unknown>[]
 }): JSX.Element => {
-  const [filter] = useQuery(['from'])
-  const [query] = useQuery()
-  usePrintArticles({
-    filter,
-    id: query.id as string
+  const [filter] = useQuery(['from', 'printFlow', 'workflowState'])
+  const [{ page }] = useQuery()
```
Victor-Lindh, line 32: Maybe I don't understand correctly, but isn't `page` also found on the `filter` variable on the line above, thus making the second call to `useQuery()` redundant...?

sinusbrus: yes, seems to work. nice catch!
and thinking one more time on this, I don't even think i need `page`

will fix!

sinusbrus: fixat och commitat

Victor-Lindh (approved):
✅

PR #1458, opened 2025-05-21
Approvals time fixes

Victor-Lindh:
- fix: New articles doesn't have a "time" in Dagen
- fix: sort column cards on publish vs start times

### DIFF
--- a/src/views/Approvals/ApprovalsCard.tsx
+++ b/src/views/Approvals/ApprovalsCard.tsx
@@ -34,7 +34,10 @@ export const ApprovalsCard = ({ assignment, isSelected, isFocused, status }: {
   const openType = (assignmentType: string) => assignmentType === 'core/flash' ? openFlash : openArticle
   const time = assignment.data.publish
     ? format(toZonedTime(parseISO(assignment.data.publish), timeZone), 'HH:mm')
-    : undefined
+    : assignment.data.start
+      ? format(toZonedTime(parseISO(assignment.data.start), timeZone), 'HH:mm')
+      : undefined
+
   const documentId = assignment._deliverableId
   const assignees = assignment.links.filter((m) => m.type === 'core/author' && m.title).map((l) => l.title)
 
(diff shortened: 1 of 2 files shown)


dannelundqvist:
Nested ternaries make the code hard to read and maintain and should be avoided. It would be better to refactor that logic into a helper function and then use that. Something along the lines of below.

```
{slot.items.sort(sortAssignments).map((assignment, cardN) => (
  <ApprovalsCard
    key={assignment.id}
    assignment={assignment}
    status={StatusSpecifications[assignment._deliverableStatus || 'draft']}
    isFocused={colN === focusedColumn && cardN === focusedCard}
    isSelected={
      (assignment._deliverableId && openEditors.includes(assignment._deliverableId)) ||
      openPlannings.includes(assignment._id)
    }
  />
))}
```

```
/**
 * Description of the logic here...
 */
const sortAssignments = (a, b) => {
  const aPublishTime = a.data.publish
  const bPublishTime = b.data.publish
  const aStartTime = a.data.start
  const bStartTime = b.data.start

  const aHasPublish = !!aPublishTime
  const bHasPublish = !!bPublishTime
  const aHasStart = !!aStartTime
  const bHasStart = !!bStartTime

  if (aHasPublish && !bHasPublish) return -1
  if (!aHasPublish && bHasPublish) return 1

  if (aHasPublish && bHasPublish && aPublishTime !== bPublishTime)
    return aPublishTime < bPublishTime ? -1 : 1

  if (aHasPublish && !bHasStart) return -1

  if (aHasStart && bHasStart)
    return aStartTime < bStartTime ? -1 : 1

  if (aHasStart && !bHasStart) return -1
  if (!aHasStart && bHasStart) return 1

  return 0
}

```

dannelundqvist (changes requested):
It works really good! However, deeply nested ternaries inside of a JSX component makes it hard to read and understand the logic.

PR #1461, opened 2025-05-22
Only article, flash and editorial-info should require reason.

dannelundqvist:
Also fixed values where setting KORR should be "fix" and RÄ should be "correction".


gularsson (approved):
Works great :+1:

PR #1465, opened 2025-05-22
Open published flash as read-only

Victor-Lindh:


dannelundqvist (approved):
Niceley done! Works fine.
We should also (in another issue/pr) ensure that we directly make the flash readonly when we have a flash open and publish it from the status menu.

gularsson (commented):
~I seem unable to create a prepared flash in a planning, works in main...~
Works fine now :thinking:

Victor-Lindh:
> ~I seem unable to create a prepared flash in a planning, works in main...~ Works fine now 🤔

Yeah, been a while since this diverged from main, going to merge the latest from there and see how it holds up 🤞🏻

gularsson (approved):
:+1:

PR #1466, opened 2025-05-22
Wait for collaborative document to be synced

dannelundqvist:
When creating new hocuspocuswebsockets to use with ad hoc collaborative documents we have to wait for the new hocuspocus connection to have its first sync. Otherwise we end up trying to create flashes and articles based on a wire before the hocuspocus is ready.


src/hooks/useCollaborationDocument.tsx
```diff
@@ -18,108 +18,64 @@ interface UseHocusPocusDocumentResult {
...
-          document: provider.document,
-          loading: false
-        }))
+        setDocument(provider.document)
+        setSynced(true)
+        setLoading(false)
       },
       onDisconnect: () => {
-        setState((prevState) => ({ ...prevState, synced: false }))
-        provider.destroy()
```
Victor-Lindh, line 91: Don't we still need to call destroy() on disconnecting...? 🤔

dannelundqvist: It shouldn't be necessary as it will be done in the cleanup(). Currently we don't recover from this state anyway, and it will probably be rendered deprecated by the changes we do now... this PR was premature... sorry.

gularsson (approved):

PR #1467, opened 2025-05-23
use env instead of hardcoded path

Victor-Lindh:


src/views/Wires/index.tsx
```diff
@@ -112,9 +112,10 @@ function handleClose(
   viewId?: string,
   history?: HistoryInterface): void {
   if (viewId && history) {
+    const BASE_URL = import.meta.env.BASE_URL
     const newContentState = (history.state?.contentState.filter((obj) => obj.viewId !== viewId) || [])
     // TODO: Get new url
```
gularsson, line 117: :broom:

src/views/Wires/index.tsx
```diff
@@ -112,9 +112,10 @@ function handleClose(
   viewId?: string,
   history?: HistoryInterface): void {
   if (viewId && history) {
+    const BASE_URL = import.meta.env.BASE_URL
```
gularsson, line 115: Maybe move to top, ie line 19ish

gularsson (approved):

PR #1470, opened 2025-05-23
events: query by timezoned date, planning created from fullday event ends up on wrong day

Victor-Lindh:


gularsson (approved):
Works nice imo

PR #1487, opened 2025-05-26
fix: flashes add section when existing planning

gularsson:


dannelundqvist (approved):
Good fix! Works as expected!

Victor-Lindh (approved):
Looks good, works as expected 👍🏻

PR #1492, opened 2025-05-27
Feature/elelog 7 godkandaapproved green articles show up as draft in dagen

dannelundqvist:


gularsson (approved):
Looks good, works good :+1:

PR #1499, opened 2025-05-30
Feature/elelog 28 endpoint for reverting document to repository version

dannelundqvist:
Also:
* fixed bug that prevented us from nesting api paths after param in path
* moved document manipulations to ber underneath /api/documents[id] to be more consistent (almost REST-ish...)


src-srv/api/documents/[id]/addassignment.ts
```diff
@@ -1,36 +1,35 @@
...
+  const planningId = req.params.id
   const locals = res.locals as Record<string, unknown> | undefined
   const session = locals?.session as { accessToken?: string, user?: Context['user'] } | undefined
 
-  if (!session?.accessToken || !session?.user) {
-    return {
-      statusCode: 401,
-      statusMessage: 'Unauthorized: Session not found, can not snapshot document'
-    }
+  const context = getContextFromValidSession(session as unknown)
```
Comment (gularsson) [inline src-srv/api/documents/[id]/addassignment.ts:26]: Not needed to cast as `unknown`?

src-srv/lib/getContextFromValidSession.ts
```diff
@@ -0,0 +1,40 @@
+import { assertContext, type Context } from './assertContext.js'
+
+export const getContextFromValidSession = (session: unknown, agent: 'server' | {} & string = 'server'): Context | {
```
gularsson, line 3: I don't really understand the agent part...
We take arg `agent` which can only be `server`and if it's not, we set it to `server`.
Why take an arg then?

src-srv/lib/getContextFromValidSession.ts
```diff
@@ -0,0 +1,40 @@
...
+  accessToken: string
+  user: Context['user']
+} => {
+  return !!value
+    && typeof value === 'object'
+    && 'accessToken' in value
+    && typeof value.accessToken === 'string'
+    && 'user' in value
+    && !!value.user
+}
```
gularsson, line 40: Whats the difference between `isSession` and `assertContext`? :thinking:

src-srv/api/documents/[id]/restore.ts
```diff
@@ -0,0 +1,88 @@
...
+ * state somehow gets corrupt.
+ *
+ * /api/documents/[uuid]/restore[?version=[nr]]
+ */
+export const POST: RouteHandler = async (req: Request, { collaborationServer, cache, repository, res }) => {
+  const id = req.params.id
+  const version = Number(req.query.version || '0')
+  const { session } = res.locals
+
+  const context = getContextFromValidSession(session as unknown)
```
Comment (gularsson) [inline src-srv/api/documents/[id]/restore.ts:24]: Cast as `unknown` not needed?

src-srv/api/documents/[id]/snapshot.ts
```diff
@@ -31,24 +32,9 @@ export const GET: RouteHandler = async (req: Request, { collaborationServer, cac
...
-    }
-  }
-
-  const context: Context = {
-    accessToken,
-    user,
-    agent: 'server'
-  }
-
+  const context = getContextFromValidSession(session as unknown)
```
Comment (gularsson) [inline src-srv/api/documents/[id]/snapshot.ts:35]: `unknown` not needed

gularsson (approved):

gularsson (changes requested):
Didn't mean to approve, just comment :slightly_frowning_face: 
Will request changes, and we'll discuss.

src-srv/lib/getContextFromValidSession.ts
dannelundqvist: The param agent is defaulted to be 'server' but can be any string as defined by `agent: 'server' | {} & string = 'server'`. The weird construct is there to allow typescript to suggest 'server' but still allow any string. We use the agent 'client' from the client. I agree that this function is not used with any other value than default, but could be.

PR #1510, opened 2025-06-05
feat: Custom print roles

vilindh:
* Companion/needs PR: https://github.com/ttab/textbit-plugins/pull/427

- Use textbit plugin PrintText to add a `tt/print-text` block to print articles, that can have a custom named role, that is to be handled specifically.

### DIFF
--- a/src/hooks/index/lib/assignments/fetchAssignments.ts
+++ b/src/hooks/index/lib/assignments/fetchAssignments.ts
@@ -1,5 +1,6 @@
 import type { Index } from '@/shared/Index'
 import type { Repository } from '@/shared/Repository'
+import type { Document } from '@ttab/elephant-api/newsdoc'
 import { QueryV1, BoolQueryV1, TermQueryV1, RangeQueryV1 } from '@ttab/elephant-api/index'
 import type { DocumentMetrics, StatusOverviewItem } from '@ttab/elephant-api/repository'
 import { type BulkGetResponse, type GetMetricsResponse, type GetStatusOverviewResponse } from '@ttab/elephant-api/repository'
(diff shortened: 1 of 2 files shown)


shared/transformations/newsdoc/index.ts
```diff
@@ -48,6 +48,7 @@ export function slateToNewsDoc(elements: TBElement[]): Block[] | undefined {
   return elements.map((element: TBElement) => {
     switch (element.type) {
       case 'core/text':
+      case 'tt/print-text':
```
gularsson, line 51: I think we should make these two separate transform functions (core/text` and `tt/print-text`), I recon it's clearer that way and the functionality isn't overlapping that much.

src-srv/utils/extensions/Snapshot.ts
```diff
@@ -106,7 +106,7 @@ export async function snapshot(uuid: string, context: Context): Promise<Snapshot
     }
 
     const base = `${process.env.PROTOCOL || 'http'}://${process.env.HOST || 'localhost'}${process.env.PORT ? `:${process.env.PORT}` : ''}`
-    const url = new URL(`${BASE_URL}/api/snapshot/${uuid}`, base)
+    const url = new URL(`${BASE_URL}/api/documents/${uuid}/snapshot/`, base)
```
gularsson, line 109: This will have to be removed, since we revert the commit with that functionality. Oh yeah, we also need to revert the commit from the feature branch.

gularsson (changes requested):

Victor-Lindh:
Moved this to https://github.com/ttab/elephant-chrome/pull/1521 instead for clarity

PR #1515, opened 2025-06-05
Assignment times should be consistently displayed

Victor-Lindh:


src/components/DataItem/AssignmentTimeDisplay.tsx
```diff
@@ -1,18 +1,27 @@
...
-type Time = { time: Array<Date | string | undefined>, tooltip: string } | undefined
+type Time = { time: Array<Date | string | undefined>, tooltip: string, type: string } | undefined
 
 export const AssignmentTimeDisplay = ({ date, className, icon }: { date: Time, className?: string, icon: React.FC<LucideProps> }): JSX.Element => {
   const { locale, timeZone } = useRegistry()
-
   const Icon = icon
 
+  function displayDay(time: Array<Date | string | undefined>): string {
+    const timeValue = typeof time[0] === 'object' ? time[0] : typeof time[0] === 'string' ? new Date(time[0]) : ''
```
gularsson, line 13: Especially since this is encapsulated in a function, I think it would be more readable not as a nested ternary.

gularsson (approved):

src/components/DataItem/AssignmentTimeDisplay.tsx
Victor-Lindh: Should be fixed in new commit

gularsson (approved):

PR #1523, opened 2025-06-09
Manyminorimprovements

sinusbrus:


src/views/PrintEditor/index.tsx
```diff
@@ -53,7 +54,7 @@ import { snapshot } from '@/lib/snapshot'
 
 const meta: ViewMetadata = {
   name: 'PrintEditor',
-  path: `${import.meta.env.BASE_URL || ''}/print-editor`,
+  path: `${import.meta.env.BASE_URL || ''}/print`,
```
gularsson, line 57: This will lead to two views named `print`, weird enough it seems to work... :thinking: 
I looked into this, if you copy and paste a url our navigation cannot handle PrintEditor (two caps) or print-editor as path. We cannot from path figure out component.
Best would be to name it something else, one word, no hyphen.

src/views/PrintEditor/components/Notes/index.tsx
gularsson, line 1: Instead of duplicating we should probably move the note-component to components/
Or are there any changes in usage?

src/views/PrintArticles/PrintArticlesList.tsx
```diff
@@ -35,6 +36,10 @@ export const PrintArticleList = ({ columns }: {
     query: constructQuery(filter),
     size: 1000,
     fields,
+    sort: [
+      SortingV1.create({ field: 'document.rel.flow.title.sort', desc: false }),
+      SortingV1.create({ field: 'document.title.sort', desc: false })
```
gularsson, line 41: document.title.sort is added as default to all queries, shouldn't be necessary.

gularsson (changes requested):

src/views/PrintEditor/index.tsx
sinusbrus: Yeah, I've noticed the problem
And even more View are using the same identifier
could we move this one to a separate PR?
https://linear.app/tt-nyhetsbyran/issue/DEV-1631/sidan-kraschar-nar-man-surfar-direkt-till-elephantprint

src/views/PrintArticles/PrintArticlesList.tsx
sinusbrus: Allright, I did it like this to get clearer grasp on how sorting i working.
We have to tweak the sorting even more, and start to show TV titles as well

But maybe I can take my new knowledge in consideration when improving the sorting?

gularsson (approved):

PR #1528, opened 2025-06-10
fix(auth): unify session and JWT auth, add jose v6

gularsson:
- Add `jose` v6 as a dependency and update usage to leverage its JWT verification features.
    - Refactor `assertAuthenticatedUser` middleware to support both Auth.js session and JWT-based authentication using a remote JWKS.
    - Introduce `getSession` utility to extract session or JWT info from requests, simplifying access to user and token data.
    - Update document API routes to use the new session extraction logic, improving consistency and security.
    - Fix snapshot extension to use correct snapshot endpoint URL.


Victor-Lindh (approved):
👍🏻 Seems to work fine

PR #1529, opened 2025-06-10
Add tv-channels search with schema, use external combobox in TVListing plugin

Victor-Lindh:
- Adds external Combobox to existing TVListing plugin to enable searching for tv channels


gularsson (approved):
:+1:

PR #1544, opened 2025-06-18
fix: remove version param from saveDocument calls

gularsson:
Fixed the saveDocument method in Repository and all its usages to remove the version parameter. The version is now handled internally within the save operation. Updated related function signatures and callers to reflect this change.


Victor-Lindh (approved):
✅

PR #1578, opened 2025-06-24
prevent certain hotkey usage in flash editors

Victor-Lindh:


Victor-Lindh:
> We need to prevent heading-1 also, otherwise you could set heading-1 for the body text

How would we then handle if the user accidentally removes the heading-1?
I do suppose that should be a restriction of its own.

gularsson (approved):

dannelundqvist (approved):

PR #1593, opened 2025-06-26
Notes: add ellipsis menu instead of closing X icon

Victor-Lindh:
, as it can cause confusion when placed close to the closing X of an article


gularsson (approved):
:+1:

dannelundqvist (changes requested):
Codewise nice, but it seems I still have an issue in Chrome.

dannelundqvist (approved):

PR #1661, opened 2025-08-01
Fix broken duplicate events function , add duplication to planning-items

vilindh:


dannelundqvist (changes requested):
Unfortunately I could not get it to work. Nothing was copied despite signaling that it was successfully copied. The link in the toast led to an empty document and it was not possible to find it by navigating to that date in the events list.

Skärmavbild 2025-08-04 kl  08 56 33


Additionally the dates should be in iso format (see individual comments).

Skärmavbild 2025-08-04 kl  08 46 41
Skärmavbild 2025-08-04 kl  08 55 43

dannelundqvist (approved):

PR #1680, opened 2025-08-11
add support for a separation of cluster-internal and public URLs

hugowetterberg:
It's more efficient to communicate directly to internal services rather than through the public domain and load balancer that clients have to go through.

This adds support for specifying a separate `[SERVICE]_PUBLIC_URL` that will be preferred over the normal URL by the "/api/urls" endpoint.

Also made the `getServerurls()` method more "data driven". Probably less typescript friendly, which I guess was the main driver for the previous method, but gives us the benefit of better error messages when a server URL is missing.

### DIFF
--- a/src/lib/getServerUrls.ts
+++ b/src/lib/getServerUrls.ts
@@ -38,9 +38,10 @@ export async function getServerUrls(): Promise<ServerUrls> {
       urls[field] = new URL(value)
     }
 
-    urls['repositoryEventsUrl'] = new URL('/sse', urls['repositoryUrl'])
-
-    return urls as unknown as ServerUrls
+    return {
+      ...urls,
+      repositoryEventsUrl: new URL('/sse', urls['repositoryUrl'])
+    } as ServerUrls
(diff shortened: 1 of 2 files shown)


src/lib/getServerUrls.ts
```diff
@@ -20,42 +20,27 @@ export async function getServerUrls(): Promise<ServerUrls> {
...
-      userUrl: new URL(userUrl),
-      faroUrl: new URL(faroUrl),
-      baboonUrl: new URL(baboonUrl)
+      if (typeof value != 'string' || value == '') {
+        throw new Error(`missing '${field}' server URL`)
+      }
+
+      urls[field] = new URL(value)
     }
+
```
dannelundqvist, line 40: For improved type safety typescript can better infer the actual keys in the struct, and make the unsafe unknown cast unnecessary, if you change these lines to:
```
return {
  ...urls,
  repositoryEventsUrl: new URL('/sse', urls['repositoryUrl'])
} as ServerUrls
```

PR #1688, opened 2025-08-15
fix: individually set event.data fields. If not, it messes up the Yjs…

Victor-Lindh:
… structure and causes an error that prevented event data to be saved


dannelundqvist (approved):
👍

PR #1691, opened 2025-08-18
Ensure correct height of app header and main menu header. Refactored …

dannelundqvist:
…main menu structure to simplify changes.

Restructed/separated main menu code into more components to separate the main menu sheets and menu items opening in "sub" sheets. (Now the main menu closes as soon as the sub sheet is opened, i.e latest published news items).

Made sure the main menu header aligns with the app header the same way regardless of which view is opened.

### DIFF
--- a/src/components/App/Menu.tsx
+++ b/src/components/App/Menu.tsx
@@ -1,5 +1,4 @@
 import {
-  Button,
   Sheet,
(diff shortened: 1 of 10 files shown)


Victor-Lindh (approved):
👍🏻

PR #1692, opened 2025-08-19
Feature/elelog 233 softcrop through focus point in images

dannelundqvist:


shared/transformations/newsdoc/core/softcrop.ts
gularsson, line 1: Maybe a simple test, works as documentation. Expected input => expected output. Should be close to copy paste from other transformations.

shared/transformations/newsdoc/core/softcrop.ts
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  const properties: Record<string, string> = {}
+  if (typeof block?.data.crop === 'string') {
+    properties.crop = block.data.crop
+  }
+
+  if (typeof block?.data.focus === 'string') {
+    properties.crop = block.data.focus
+  }
```
gularsson, line 25: Shouldn't this be be `properties.focus`?

gularsson (changes requested):

PR #1693, opened 2025-08-19
fix(table): handle empty table rows as error objects

gularsson:
Previously, empty rows in tables were not handled, which could lead to unexpected behavior or display issues. This change ensures that empty rows are returned as 'core/table/row/error' objects, allowing them to be displayed as 'unknown object' and improving robustness.


dannelundqvist:
Works well!

An alternative solution, if we know how many cells there should be would be to create an empty row. Like below. But it might be better to actually deem empty rows an error.

```
{
            type: 'core/table/row/cell',
            class: 'text',
            children: [{ text: '' }],
            properties: {
              rowspan: 4
            }
}
```

PR #1708, opened 2025-08-21
Assignments: list view should show status of all deliverables

vilindh:


src/components/Commands/FacetedFilter.tsx
```diff
@@ -53,10 +54,17 @@ export const FacetedFilter = <TData, TValue>({
...
-          )}
-          <span>{option.label}</span>
+          <div className='flex items-center gap-2'>
+            {option.icon && (
+              <option.icon
+                size={18}
+                strokeWidth={1.75}
+                color={DocumentStatuses.find((status) => status.value === option.value)?.iconProps?.color}
+                className={DocumentStatuses.find((status) => status.value === option.value)?.iconProps?.className}
+              />
```
gularsson, line 64: This is a generic component, using `DocumentStatuses` here works now since we're not using colors for any other filter option.

I think we can just do:
```suggestion
              <option.icon
                size={18}
                strokeWidth={1.75}
                color={option.iconProps?.color}
                className={option.iconProps?.className}
              />
```
Since `options` should be declared on the column definition. As defined here: https://github.com/ttab/elephant-chrome/pull/1708/files#diff-c85496fd9266e099f4b6aaea3a4d7c2716dbde1d0782840b89ea7af91bae718dR38

src/hooks/index/useDocuments/lib/fetch.ts
```diff
@@ -56,10 +62,35 @@ export async function fetch<T extends HitV1, F>({
 
   let result = hits
 
-  // Format planning result as assingments
-  if (options?.asAssignments && query) {
+  // Format planning result as assignments
+  if (options?.asAssignments && query && repository) {
     // FIXME: Could this be better
```
gularsson, line 67: Maybe still relevant, however it applies to a larger chunk of code, not just the type coercion on the return. :broom:?

src/hooks/index/useDocuments/lib/fetch.ts
```diff
@@ -56,10 +62,35 @@ export async function fetch<T extends HitV1, F>({
...
+        all.push(...deliverables)
+      }
+      return all
+    }, []).flat()
+
+    const statuses = await repository.getStatuses({
+      uuids,
+      statuses: knownStatuses,
+      accessToken: session.accessToken
+    })
```
gularsson, line 92: The statuses could preferably be extracted to a function.
```
    return asAssignments(result as unknown as Assignment[], query, getStatuses(<args>) as unknown as T[]

gularsson: placed in `src/hooks/index/useDocuments/lib`?

gularsson (changes requested):
Works fine! Some suggestions

gularsson (approved):

PR #1721, opened 2025-08-27
refactor(notes): deduplicate and centralize Notes components

gularsson:
Moved Notes and AddNote components to a shared location under src/components/Notes, removing duplicated implementations from PrintEditor. Updated all imports to use the centralized components. This simplifies maintenance and ensures consistent note handling across editors.


Victor-Lindh (approved):
👍🏻

PR #1723, opened 2025-08-27
Reflect article status change in assignment row

Victor-Lindh:


dannelundqvist:
Nice! I'd like to see a few minor tweaks so it does not get messy to the eye as we are adding more information.

Skärmavbild 2025-09-01 kl  10 30 17

PR #1728, opened 2025-08-28
Approvals: show version number and cause in card

Victor-Lindh:


gularsson (approved):
Nice :+1:
I don't know if we want to show v1, just if version > 1... might come feedback for this.

PR #1731, opened 2025-08-29
Approvals card byline improvements

Victor-Lindh:
Refined logic for outputting names on Approval view cards

### DIFF
--- a/src/views/Approvals/AuthorNames.tsx
+++ b/src/views/Approvals/AuthorNames.tsx
@@ -18,7 +18,7 @@ export const authorOutput = (matchedAuthor: IDBAuthor) => {
      * 5. Whoever updates the status after that is to be output next
      *    Could be the same as the reporter, in that case don't bother outputting it again
   */
-export const makeAuthorNames = ({
+export const AuthorNames = ({
   byline,
   doneStatus,
   authors,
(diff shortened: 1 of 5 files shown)


src/views/Approvals/makeAuthorNames.tsx
dannelundqvist, line 1: As this is not a component it should have `.ts` instead of `.tsx` suffix. And probably in a `./utils` folder. If this functionality i generic and could be useful in other places, than it should be in a global utils folder, otherwise local.

dannelundqvist (changes requested):

src/views/Approvals/AuthorNames.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+     * Names are displayed using this order of priority:
+     * 1. If there is a byline, output the byline
+     * 2. If there is no byline, output the signature of each assignee
+     * 3. If there are no assignees, output the initials whoever marked the deliverable as 'done'
+     * 4. The user who changes status from draft to something else is to be output first
+     * 5. Whoever updates the status after that is to be output next
+     *    Could be the same as the reporter, in that case don't bother outputting it again
+  */
```
gularsson, line 20: There's something off with the priority... Byline for example is rarely displayed

Victor-Lindh: My mistake, got the byline value from the wrong place. Should be fixed in new commit!

PR #1735, opened 2025-09-02
fix(print): enable crop for tt/visual

gularsson:


dannelundqvist (changes requested):
It probably works just as fine as plugins overrride earlier ones. But it would still be good to remove line 112 to avoid double instantiation.

PR #1746, opened 2025-09-08
fix: use elc:: prefix for locks

gularsson:


src-srv/utils/Redis.ts
```diff
@@ -21,9 +24,9 @@ export class Redis {
   }
 
   async get(key: string): Promise<Uint8Array | undefined> {
-    const cachedDoc = await this.#redisClient?.get(`elc::hp:${key}`)
+    const cachedDoc = await this.#redisClient?.get(`${BASE_PREFIX}${key}`)
```
Victor-Lindh, line 27: : after BASE_PREFIX

Victor-Lindh (changes requested):

PR #1762, opened 2025-09-11
Feature/elelog 277 persist filters when navigating from the menu

gularsson:


dannelundqvist (changes requested):
I could not get it to remove a single section filter. It is persisted over reloads even when seemingly removed.

https://github.com/user-attachments/assets/2b79e186-7df5-4678-940a-03b049a24618

PR #1768, opened 2025-09-12
feat(approvals): add focus style to planning icon

gularsson:
Adds `md:group-focus:opacity-70` to the planning icon in ApprovalsCard so the option will be shown when navigating using keyboard.


Victor-Lindh (approved):
👍🏻

dannelundqvist (approved):
Too late in my comment.

Works really well now 🥇

PR #1776, opened 2025-09-16
Only user adding a new assignment should see the dialog. 

dannelundqvist:
This PR depends on [https://github.com/ttab/textbit/pull/318](https://github.com/ttab/textbit/pull/318) which is why I had to remove all those `// @ts-expect-error`.


src/views/Planning/components/AssignmentTable.tsx
```diff
@@ -56,7 +58,9 @@ export const AssignmentTable = ({ asDialog = false, documentId, onChange }: {
 
     appendAssignment({
       document: provider.document,
-      inProgress: true,
+      inProgress: {
+        sub: session?.user?.id ?? ''
+      },
```
gularsson, line 63: i think you ought to use the actual sub, `session.user.sub` these are stable.

gularsson (changes requested):

PR #1785, opened 2025-09-18
fix(nav): allow alt+arrow navigation in with editor focus

gularsson:
Permit alt+left and alt+right key navigation even when focus is in an editor view. This enables custom view navigation shortcuts without browser interference, while preserving default arrow key behavior for text editing.

### DIFF
--- a/src/hooks/useNavigationKeys.tsx
+++ b/src/hooks/useNavigationKeys.tsx
@@ -54,6 +54,12 @@ export const useNavigationKeys = (
 
     // Don't override arrow navigation in editable targets
     if (target && isEditableElement(target)) {
+      // But allow alt+left and alt+right, for view navigation.
+      if (event.altKey && (event.key === 'ArrowLeft' || event.key === 'ArrowRight')) {
+        // Browser navigation default will intercept otherwise
+        event.preventDefault()
+        onNavigation(event)
+      }
       return
     }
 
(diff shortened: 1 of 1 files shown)


dannelundqvist:
Can't really agree with this one. `alt+→` and `alt+←` is used to jump between words in almost any text editor. This PR still maintains that but at the same time the focused view changes. But without focus being changed from the editor.

dannelundqvist:
> Well, it doesn't do that in _our_ editor... alt+left/right gets caught by the browser and thats browser default for navigating in the browser history. Since we've hi-jacked that in all other places, I figured why not here also?

It does that in Chrome on OS X at least. Giving me this somewhat confusing UX. Seems we need to investigate behaviour in more browsers/OSs...

src/hooks/useNavigationKeys.tsx
```diff
@@ -103,3 +112,23 @@ function isEditableElement(element: Element | null): element is HTMLInputElement
...
+
+  if (!platform && typeof navigator.platform === 'string') {
+    platform = navigator.platform
+  }
+
+  if (!platform) {
+    platform = navigator.userAgent
+  }
+
+  return /\bmac\b/i.test(platform)
```
dannelundqvist, line 133:
```suggestion
  return /\bmac/i.test(platform)
```

dannelundqvist (changes requested):
The regex needs to be more open as you often get "macOS".

PR #1821, opened 2025-10-02
fix(navigation): apply query to correct view when keepfocus is used

gularsson:
- Fix link handling to avoid duplicate query strings in path
- Remove unused `activeId` prop from `GroupedRows` and related logic
- Use `openDocuments` to determine selected rows in `Row` component
- Update row styling to use `data-state=selected` for selection
- Add 'PrintEditor' to open document names in `useOpenDocuments`
- Refactor `DateChanger` for better date handling and link props


dannelundqvist (approved):
Nice fixes 👍

PR #2384, opened 2026-03-28
Feature/elelog 335 wires improved search capability

gularsson:
feat(AdvancedSearch): add advanced search dialog for SearchOverview and Wires
    
Introduce a structured and query-syntax advanced search mode with
i18n support (en/nb/sv), branded types for field paths, runtime
state validation at JSON boundaries, and URL param serialization.
    
Badges display in the toolbar with tooltip details and open the
edit dialog on click. Responsive search bar width scales with
viewport. Includes tests for buildQuery, summarize, and
serialize/deserialize roundtrip.


vilindh (changes requested):
Looks nice, just a few suggestions of improvement!

vilindh (approved):
Looking good!

LillaBlond:
It seems nothing show up if I add a date intervall. For example if I search for Trump I can see wires with Trump in the last few days. If I add a search intervall for the same time period I don't get any results.

For the dates maybe it shouldn't be possible to add only from or to date but if user press search only one of them are selected the other will automatically change to today.

Would also be nice if there is butto/text/icon etc that allow clearing both date filters in one go instead of having to click on them and remove one by one.

LillaBlond:
If I choose exact fras och fuzzy match at the same time and try to search I end up in an error state

Skärmavbild 2026-04-16 kl  10 59 54

LillaBlond (changes requested):
Please see comments above for adding date filter and fuzzy match combined with exact phrase

PR #2389, opened 2026-03-30
Feature/elelog 479 scope the publish prompt to relevant container

LillaBlond:
This pr transforms the display of the prompt message.
Changes:

- Highlights the background of the document being actioned slightly compared to the rest of site which have a darker background when the prompt message shows up.
- Adds icon for the relevant type in the prompt message
- Specify what document type is being actioned in the prompt message eg. Publish article or Publish planning
- Add translation strings for the prompt title for different types instead of generic title.


src/components/DocumentStatus/PromptDefault.tsx
```diff
@@ -3,14 +3,17 @@ import { Prompt } from '../Prompt'
 import { useCallback, useEffect, useState } from 'react'
 import { PromptCauseField } from './PromptCauseField'
 import { useTranslation } from 'react-i18next'
+import type { LucideIcon } from 'lucide-react'
```
vilindh: Should import from @ttab/elephant-ui/icons

vilindh (changes requested):

gularsson (changes requested):
We should use `documentTypeValueFormat` instead of hard coding the icons in respective header.

This would also reduce/remove the use for sending icon as a prop.

vilindh (approved):

gularsson (approved):

PR #2414, opened 2026-04-08
Feature/elelog 81 not possible to schedule some articles

LillaBlond:
- Add the possibility to change date when scheduling articles
- Removing schedule options from flashes
- Show the scheduled date and time on the assignment for when scheduled
- Can choose if move a scheduled article or not when moving planning 
- Scheduled article will show up in planing for the date and time it's scheduled for.


src/views/Assignments/lib/getAssignmentStatuses.tsx
dannelundqvist, line 1: Optional chaining does not handle if properties exist but contains no items: add `?.` before `.uuid` and `.value` on every line.

dannelundqvist (changes requested):

src/views/Assignments/lib/getAssignmentStatuses.tsx
LillaBlond: Thanks for highlighting this. I've changed it.

src/views/Assignments/lib/getAssignmentStatuses.tsx
```diff
@@ -0,0 +1,44 @@
...
+    const result = await repository?.getStatuses({ uuids: assignmentsUUIds, statuses: acceptedStatues, accessToken: accessToken })
+    if (!result) {
+      return undefined
+    }
+    return result.items.map((assignment) => {
+      return { id: assignmentsId.find((item) => item?.uuid === assignment.uuid)?.id, uuid: assignment.uuid, workflowState: assignment.workflowState }
+    })
+  } catch (error) {
+    console.log(error)
+  }
```
gularsson, line 45: we need to display this to the user

src/views/Assignments/lib/getAssignmentStatuses.tsx
```diff
@@ -0,0 +1,44 @@
+import type { Repository } from '@/shared/Repository'
+import type { EleBlock } from '@/shared/types'
+
+export const getAssignmentStatuses = async (assignments: EleBlock[], repository: Repository, accessToken: string) => {
+  const acceptedStatues = ['usable', 'withheld', 'cancelled', 'draft', 'unpublished', 'approved', 'done']
```
gularsson: typo.
is there some other place from where we could derive these values from... seems easy to forget these hard coded values.

src/views/Planning/components/MoveDialog.tsx
```diff
@@ -30,9 +32,26 @@ export const MoveDialog = ({ ydoc, onClose, newDate }: {
...
+    }
+
+    getAssignmentStatuses(assignments, repository, data?.accessToken).then((statuses) => {
+      setAssignmentStatuses(statuses)
+    }).catch((err) => {
+      onClose()
+      toast.error(t('errors:toasts.fetchAssignmentStatuses'))
+      console.error('Failed to get assignment statuses', err)
+    })
+  }, [assignments, repository, data?.accessToken, t])
```
gularsson: missing dep `onClose`

src/views/Planning/components/MoveDialog.tsx
```diff
@@ -42,56 +61,64 @@ export const MoveDialog = ({ ydoc, onClose, newDate }: {
...
-                    newDate={newDate}
-                    onChangeSelected={(id, value) => {
-                      assignmentTimes.current[id] = value
-                    }}
-                  />
-                )
-              }
-            })}
-          </div>
+          {!!assignmentStatuses && (
```
gularsson: Maybe a loader here, while fetching `assignmentStatuses`?

src/views/Assignments/lib/getAssignmentStatuses.tsx
```diff
@@ -0,0 +1,44 @@
...
+      default:
+        return
+    }
+  })
+
+  const assignmentsUUIds = assignmentsId.map((item) => item?.uuid).filter((uuid) => uuid !== undefined)
+
+  try {
+    if (!repository || !accessToken) {
+      return
```
gularsson, line 33: Maybe throw, catch and toast - otherwise it's a bit of a silent failure?

src/views/Assignments/lib/getAssignmentStatuses.tsx
gularsson, line 1: Not a component, should be .ts?

src/views/Assignments/lib/getAssignmentStatuses.tsx
LillaBlond: I throwed it the the new catch to be presented in a toast error

src/views/Assignments/lib/getAssignmentStatuses.tsx
LillaBlond: Took the statuses from StatusSpecifications here in stead

src/views/Planning/components/MoveDialog.tsx
LillaBlond: Added this :)

src/views/Assignments/lib/getAssignmentStatuses.tsx
LillaBlond: I throwed an error here as well

src/views/Assignments/lib/getAssignmentStatuses.tsx
LillaBlond: changed

src/locales/nb/shared.json
```diff
@@ -33,6 +33,7 @@
     "pickCause": "Velg en grunn...",
     "planningDate": "Dato i planen",
     "setTime": "Skriv inn tid",
+    "setDate": "Angi dato",
```
vilindh, line 36: For the `sv` and `en` locales, setTime and setDate have the same prefixes (Ange (x), Set (x)).
Same should be done for `nb`, to avoid inconsistencies.

LillaBlond: Fixed this :)

src/views/Assignments/lib/getAssignmentStatuses.tsx
gularsson: Great, only the typo left then ```acceptedStatues => acceptedStatuses```

src/views/Assignments/lib/getAssignmentStatuses.ts
```diff
@@ -0,0 +1,44 @@
...
+      default:
+        return
+    }
+  })
+
+  const assignmentsUUIds = assignmentsId.map((item) => item?.uuid).filter((uuid) => uuid !== undefined)
+
+  try {
+    if (!repository || !accessToken) {
+      return
```
gularsson: Still just a return :thinking:

src/views/Planning/components/MoveDialog.tsx
gularsson: Revisiting this I don't think we can have `onClose` in deps, the dep will cause the effect to retrigger on every rerender.

so way forward is probably `// eslint-disable-next-line react-hooks/exhaustive-deps` and an explanation to why we're ignoring the error...

src/views/Assignments/lib/getAssignmentStatuses.ts
vilindh: Seems to have been lost in this commit?
https://github.com/ttab/elephant-chrome/pull/2414/changes/e276a682bfb6f7cc22ad4a62a3b577faeb15a25d

src/views/Assignments/lib/getAssignmentStatuses.tsx
LillaBlond: fixed

src/views/Assignments/lib/getAssignmentStatuses.ts
LillaBlond: readed a throw here

src/views/Planning/components/MoveDialog.tsx
LillaBlond: Added this as well :)

gularsson (approved):

PR #2430, opened 2026-04-13
Feature/elelog 557 npk translation script selector dialog defaults

dannelundqvist:
User defined translation to nynorsk.
User preferences view with translation settings (and readonly view of wires columns settings).


src/hooks/useHasUnit.tsx
```diff
@@ -0,0 +1,25 @@
...
+      return false
+    }
+
+    try {
+      const decoded = decodeJwt(session.accessToken)
+      const units = Array.isArray(decoded.units) ? decoded.units as string[] : []
+      return units.includes(unit)
+    } catch {
+      return false
+    }
```
gularsson: We could, now, do:
```suggestion
    try {
      return session.includes(unit)
    } catch {
      return false
    }
```
Or skip altogether?

src/views/WireCreation/lib/translateWireContent.ts
```diff
@@ -0,0 +1,113 @@
...
+      source_language: 'nb',
+      target_language: 'nn',
+      prefs_template: 'standard',
+      ...(mode === 'personal' && personalPrefs ? { prefs: parsePersonalPrefs(personalPrefs) } : {})
+    })
+
+    if (result.texts?.values?.length === texts.length) {
+      replaceTexts(cloned, result.texts.values)
+    }
+  }
```
gularsson: If this check fails, do we return the original? Should we perhaps notify the user? Or translate what we can?

gularsson (changes requested):

src/hooks/useHasUnit.tsx
dannelundqvist: Nice! I kinda like having the hook showing the intent clearly in the code. Even if it is simple. If you strongly disagree I'll change it.

src/views/WireCreation/lib/translateWireContent.ts
dannelundqvist: True, I think we should make sure to surface errors like this.

shared/Repository.ts
```diff
@@ -24,9 +24,25 @@ import { isValidUUID } from './isValidUUID.js'
...
+  try {
+    const decoded = decodeJwt(accessToken)
+    const units = Array.isArray(decoded.units) ? decoded.units as string[] : []
+
+    if (units.includes('core://unit/redaktionen-npk')) {
+      return [{ uri: 'core://unit/redaktionen-npk', permissions: ['r', 'w'] }]
+    }
+  } catch {
+    // JWT decode failure → fall back to default ACL
+  }
```
gularsson, line 41: use session.units?

src/views/WireCreation/lib/createArticle.tsx
```diff
@@ -58,6 +99,20 @@ export async function createArticle({
...
+    try {
+      const translatedContent = await translateWireContent(wireContent, translationMode, personalPrefs)
+      if (yjsDocument) {
+        yjsDocument.getMap('ele').set('content', translatedContent)
+      }
+    } catch (ex) {
+      console.error('Translation failed, using original wire content', ex)
+      toast.error(i18n.t('wires:creation.translationError'))
+    }
+  }
```
gularsson, line 114: This could probably be expanded later on - retry or a more "in-your-face" prompt asking whether the user want to continue with the original, and not default to continue the creation process...

__tests__/buildAcl.test.ts
```diff
@@ -0,0 +1,97 @@
...
+  url: '',
+  title: 'Test',
+  content: [],
+  meta: [],
+  links: [],
+  language: ''
+}
+
+function getAclFromLastCall(): Array<{ uri: string, permissions: string[] }> {
+  return mockUpdate.mock.calls[0][0].acl
```
gularsson: would be nice to handle these lint warnings/or snooze them

shared/translate.ts
```diff
@@ -0,0 +1,35 @@
...
+    headers: {
+      'Content-Type': 'application/json'
+    },
+    body: JSON.stringify(request)
+  })
+
+  if (!response.ok) {
+    const text = await response.text().catch(() => response.statusText)
+    throw new Error(`Translation failed: ${text}`)
+  }
```
gularsson, line 33: Would be nice with a timeout here, otherwise i suspect it will hang indefinitly

src/views/WireCreation/lib/translateWireContent.ts
gularsson, line 1: Could do with some tests

gularsson (approved):

shared/Repository.ts
dannelundqvist: The session object is not always available in this context where it can be used both in srv and client.

gularsson (approved):

vilindh (approved):

PR #2449, opened 2026-04-16
Factbox view list remake

LillaBlond:
This PR change the appearance of the factbox view.

- show factboxes both published articles and factbox documents.
- Show icons for where it originates from. Article or original
- quickfilter to switch between factboxes from articles or originals.
- Also change the status flow for factbox original in discussion with Helena. It should only go from draft to Klar and if needed to removed.


src/hooks/index/useDocuments/lib/withArticleFactboxes.ts
dannelundqvist, line 1: This will fetch the last 100 (as fetch has a default of 100) embedded factboxe only. Is that intentional, or do we need to add pagination handling?

src/views/Factboxes/FactboxColumns.tsx
```diff
@@ -39,29 +41,36 @@ export function factboxColumns<Ns extends Namespace>({ locale, timeZone, t }: {
...
+        const currentStatus = row.original.fields['workflow_state']?.values[0]
+        const origin = row.original.fields['_document_origin']?.values[0] ?? 'core/factbox'
         return (
-          <DocumentStatus status={status} isChanged={isChanged} type='core/factbox' />
+          <div className='relative w-fit'>
+            {
+              origin === 'core/article'
+                ? (
+                    <div title={t('workflows:base.usable.title')}>
+                      <CircleCheckIcon strokeWidth={1.75} className='text-white rounded-full dark:text-blackbg-usable fill-usable' />
```
dannelundqvist, line 63: Missing space between `dark:text-black` and `dark:text-black`. Should be:
`text-white rounded-full dark:text-black bg-usable fill-usable`

src/views/Factbox/index.tsx
```diff
@@ -41,23 +44,60 @@ const meta: ViewMetadata = {
...
+  const embeddedMatch = typeof rawId === 'string' ? rawId.match(/^(.+):embedded:(\d+)$/) : null
+  const articleId = embeddedMatch?.[1]
+  const embeddedIndex = embeddedMatch ? parseInt(embeddedMatch[2]) : null
+  const documentId = embeddedMatch ? articleId : rawId
+
+  // If this is an embedded factbox, fetch the article and extract the factbox block
+  useEffect(() => {
+    if (!articleId || embeddedIndex === null || !repository || !session?.accessToken) return
+
+    void repository.getDocument({ uuid: articleId, accessToken: session.accessToken }).then((response) => {
```
dannelundqvist, line 63: We might need a catch and some user feedback here.

src/views/Factbox/index.tsx
```diff
@@ -41,23 +44,60 @@ const meta: ViewMetadata = {
...
-
     return toGroupedNewsDoc({
       version: 0n,
       isMetaDocument: false,
       mainDocument: '',
       subset: [],
       document: props.document || getTemplateFromView('Factbox')(documentId)
     })
-  }, [documentId, props.document])
+  }, [documentId, props.document, embeddedMatch])
```
dannelundqvist, line 100: Unless `null` the `embeddedMatch` will be a new object every render and thus the useMemo will not have the intended effect. This would be more performant and readable without the `useMemo()` call.

dannelundqvist (changes requested):

src/views/Factbox/index.tsx
LillaBlond: Addad catch with toast message here

src/views/Factbox/index.tsx
LillaBlond: removed useMemo as suggested.

src/views/Factboxes/FactboxColumns.tsx
vilindh, line 63: Document status icons differ from those in other views, Plannings for example.
See screenshots below.

Image
Image
Image
Image

src/views/Factbox/FactboxHeader.tsx
```diff
@@ -39,12 +40,14 @@ export const FactboxHeader = ({ ydoc, onDialogClose, asDialog }: {
               <ViewHeader.RemoteUsers ydoc={ydoc} />
             )}
           </div>
+          <ViewHeader.Action ydoc={ydoc} onDialogClose={onDialogClose} asDialog={asDialog}>
```
vilindh, line 43: `FactboxHeader` items should have same align as other views, Status button for example should be aligned to the right

Image
vs
Image

src/views/Factboxes/FactboxColumns.tsx
```diff
@@ -104,6 +113,31 @@ export function factboxColumns<Ns extends Namespace>({ locale, timeZone, t }: {
       },
       enableGlobalFilter: true
     },
+    {
+      id: 'documentOrigin',
+      meta: {
+        name: 'Origin',
```
vilindh, line 119: `name` should be localized

vilindh (changes requested):
Some comments about header alignment and other things in the list columns

src/views/Factboxes/FactboxColumns.tsx
```diff
@@ -104,6 +113,31 @@ export function factboxColumns<Ns extends Namespace>({ locale, timeZone, t }: {
       },
       enableGlobalFilter: true
     },
+    {
```
vilindh, line 116: Selected quick-filters are shown in the view header in other views. Here it doesn't.
Not sure it's needed though, since it's the only filter available.
Still, it's a step away from the current app convention.

src/defaults/workflowSpecification.tsx
```diff
@@ -226,10 +226,10 @@ export const getStatusSpecifications = (status: string, documentType?: string):
       }
     case 'usable':
       return {
-        icon: documentType === 'core/factbox' ? BadgeCheckIcon : CircleCheckIcon,
+        icon: CircleCheckIcon,
         className: cn(
           'text-white  rounded-full dark:text-black',
-          documentType === 'core/factbox' ? 'bg-approved fill-approved' : 'bg-usable fill-usable'
+          documentType === 'core/factbox' ? 'bg-approved fill-done' : 'bg-usable fill-usable'
```
vilindh: Should be `bg-done fill-done`

src/hooks/index/useDocuments/lib/withArticleFactboxes.ts
LillaBlond: It was intentional to not fetch that many. It's tricky to add pagination to this one as to show the factboxes correctly I had to fetch data from both the index and repository and then combine them so using the pagination setup already existing wouldn't work. 

But after this comment I've changed it now to fetch 1000 factboxdocuments and 1000 articles with embedded factboxes, combine them and sort them by modified date. Then I've used useEffect to make a temporary pagination. This is not the best solution and it needs to be changed but it adds a way to show a bit more factboxes.

Hopefully when integrating Gustav's advanced search here this could be worked on to show more correct data.

src/defaults/workflowSpecification.tsx
LillaBlond: Cheers! I fixed this

src/views/Factbox/FactboxHeader.tsx
LillaBlond: Corrected this  now :) as well as increasing the size of the factbox modal when open

src/views/Factboxes/FactboxColumns.tsx
LillaBlond: fixed that with Dannes comment it seems.

src/views/Factboxes/FactboxColumns.tsx
LillaBlond: fixed :)

src/views/Factboxes/FactboxColumns.tsx
LillaBlond: It should show so it's possible to reset filter from mobile view. However I think search and Filter functionality will have to be it's own thing. 

Together with how I need to fetch data to show this list (combination of index search and repository) it might be a bit more complicated than for example planning and events

src/views/Factbox/FactboxHeader.tsx
vilindh: There is still some diff in alignment. See screenshots below, compared to the Planning view.
I see there is a max-w-[750] on the div container, I don't think that's needed here.

Skärmavbild 2026-04-20 kl  09 25 20
Skärmavbild 2026-04-20 kl  09 25 26

LillaBlond: ok now it should be the exact same space between. gap-4

dannelundqvist (approved):

vilindh (approved):

PR #2455, opened 2026-04-17
Feature/elelog 564 timeless articles new document variant and dedicated view

gularsson:


shared/templates/timelessArticleDocumentTemplate.ts
```diff
@@ -0,0 +1,55 @@
+import { Block, Document } from '@ttab/elephant-api/newsdoc'
+import type { TemplatePayload } from './index.js'
+import { getSystemLanguage } from '@/shared/getSystemLanguage.js'
+
+export function timelessArticleDocumentTemplate(
+  id: string,
+  payload?: TemplatePayload
+): Document {
+  delete payload?.meta?.['core/description']
```
dannelundqvist, line 9: Mutates caller´s payload. Fine if intentional or of no consequence. Otherwise maybe we should make a shallow copy first?

dannelundqvist:
Ideally we should stick to the UI patterns we have established. Attached create planning as example.
Skärmavbild 2026-04-21 kl  11 31 13
Skärmavbild 2026-04-21 kl  11 30 47

dannelundqvist (approved):

vilindh (approved):
Great work!

shared/templates/timelessArticleDocumentTemplate.ts
gularsson: It's the same pattern as used in `articleDocumentTemplate` - payload is sourced from planning and we don't want `core/description` appended to the timeless.

PR #2472, opened 2026-04-21
Factbox from original remake

LillaBlond:
When opening an original factbox this PR adds a list of articles this factbox is used in at the bottom.


src/locales/nb/factbox.json
```diff
@@ -2,10 +2,14 @@
...
   },
+  "usedInArticles": "Brukes i artikler",
   "prompts": {
     "approveBeforeClose": "Vil du godkjenne endringene før du lukker?",
     "savedButNotApproved": "Endringene er lagret, men faktaboksen er ikke godkjent."
-  }
-}
\ No newline at end of file
+  },
+  "lockedToArticle": ""
```
dannelundqvist: Is the Norwegian label intentionally blank?

src/views/Factbox/FactboxHeader.tsx
```diff
@@ -29,20 +44,50 @@ export const FactboxHeader = ({ ydoc, onDialogClose, asDialog }: {
...
-            )}
+            <>
+              {!asDialog && ydoc
+                ? (
+                    <StatusMenu
+                      ydoc={ydoc}
+                    />
+                  )
+                : (
+                    <div className='bg-zink-50 dark:bg-slate-700 text-xs py-1 px-3 border-2 rounded border-slate-200'>
```
dannelundqvist: Should maybe be `bg-zinc-50`? There is no tailwind `zink` color.

dannelundqvist (changes requested):
Two very small changes. Other than that it's great!

src/locales/nb/factbox.json
LillaBlond: is was a duplicate string, removed now

src/views/Factbox/FactboxHeader.tsx
LillaBlond: this is removed completely as this element is not in the header anymore

vilindh (approved):
Good job!

PR #2486, opened 2026-04-23
Editor: prevent block class actions from being displayed and added fr…

vilindh:
…om gutter content menu

### DIFF
--- a/src/components/Editor/ContentMenu.tsx
+++ b/src/components/Editor/ContentMenu.tsx
@@ -7,6 +7,10 @@ export const ContentMenu = (): JSX.Element => {
   const editor = useEditor()
   const { actions } = usePluginRegistry()
 
+  // Check in what plugin context the user is upon selecting from the menu.
+  // If they are inside a block class element, other block class elements should not be selectable to insert,
+  // as we currently have no need for adding one block class element inside another.
+  // However, this might change in future iterations.
   const currentPluginClass = (() => {
     if (!editor.selection) {
       return null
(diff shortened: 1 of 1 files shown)


dannelundqvist (commented):
Worth noting is that this will effectively stop us from adding images to a factbox if we want to later. For now that is fine as we don't allow images in factboxes anyway.

src/components/Editor/ContentMenu.tsx
```diff
@@ -1,13 +1,31 @@
...
 
 export const ContentMenu = (): JSX.Element => {
+  const editor = useEditor()
   const { actions } = usePluginRegistry()
 
+  // Check in what plugin context the user is upon selecting from the menu.
+  // If they are inside a block class element, other block class elements should not be selectable to insert,
+  // as we currently have no need for adding one block class element inside another.
+  // However, this might change in future iterations.
+  const currentPluginClass = (() => {
```
dannelundqvist: Could this code not be replaced by importing TextbitElement and use the helper there?

```
  const topNode = editor.selection
    ? editor.children[editor.selection.anchor.path[0]]
    : undefined
  const insideBlock = TextbitElement.isBlock(topNode)
```

dannelundqvist (changes requested):

src/components/Editor/ContentMenu.tsx
vilindh: Good catch, I switched to using the native method instead!

dannelundqvist (approved):

PR #2494, opened 2026-04-24
No validaton of sluglines uniqueness, no slugline on planning item it…

dannelundqvist:
…self.

### DIFF
--- a/src/lib/getServerEnvs.ts
+++ b/src/lib/getServerEnvs.ts
@@ -72,7 +72,8 @@ export async function getServerEnvs(): Promise<ServerConfig> {
       },
       featureFlags: {
         hasPrint: data['hasPrint'] ? !!data['hasPrint'] : false,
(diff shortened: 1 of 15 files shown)


src/lib/getServerEnvs.ts
```diff
@@ -72,7 +72,8 @@ export async function getServerEnvs(): Promise<ServerConfig> {
       },
       featureFlags: {
         hasPrint: data['hasPrint'] ? !!data['hasPrint'] : false,
-        hasHast: data['hasHast'] ? !!data['hasHast'] : false
+        hasHast: data['hasHast'] ? !!data['hasHast'] : false,
+        hasLooseSlugline: data['hasLooseSlugline'] ? !!data['hasLooseSlugline'] : false
```
gularsson, line 76: Not for this PR, but we really should fix this. `HAS_LOOSE_SLUGLINE=false` will be coerced to `true`

gularsson (approved):

PR #2503, opened 2026-04-28
Factboxes remade

vilindh:


src/views/Factboxes/FactboxList.tsx
```diff
@@ -1,35 +1,45 @@
...
+  useEffect(() => {
+    if (!data) return
+    // TO FIX: added pagination on the client side as a temporary solution. Needs to be reworked to show more correct data.
+    const start = (currentPage - 1) * PAGE_SIZE
+    setData(data.slice(start, start + PAGE_SIZE))
+  }, [data, currentPage, setData])
+
   const onRowSelected = useCallback((row?: Factbox) => {
     if (row) {
       console.info(`Selected planning item ${row.id}`)
```
dannelundqvist: I think we should remove this (even though it is not part of this PR). Unnecessary and not a planning item either.

shared/transformations/newsdoc/core/lists.ts
dannelundqvist, line 1: Not extremely important. But all other files in `./core` all use camel case (starting with lowercase).

dannelundqvist:
As we have added "Skapa faktaruta", should we use the term "Sök faktarutor" to follow the pattern from "Sök bilder"? Or use only "Bilder"?

Skärmavbild 2026-04-29 kl  09 16 59

dannelundqvist:
Factbox header is not localized in readonly view (in published article). And it does not contain the same information as in the editable view (should it?).

Readonly view
Skärmavbild 2026-04-29 kl  09 24 53

Editable view
Skärmavbild 2026-04-29 kl  09 25 03

dannelundqvist (approved):

shared/transformations/newsdoc/core/lists.ts
vilindh: Done

src/views/Factboxes/FactboxList.tsx
vilindh: Done!

PR #2511, opened 2026-04-28
page query param should reset correctly

vilindh:


gularsson (approved):
I think `src/components/Filter/BaseSelected.tsx` and `src/views/SearchOverview/Commands.tsx` could do with these changes also?

gularsson (approved):

PR #2513, opened 2026-04-29
Buttongroup item box click surface

vilindh:
Makes sure the click surface of each dropdown item in the AddButtonGroup component is the entire item container - not just the text surface.

### DIFF
--- a/src/components/Header/AddButtonGroup.tsx
+++ b/src/components/Header/AddButtonGroup.tsx
@@ -45,8 +45,6 @@ const AddButton = ({
   onClick: (view: ButtonView) => void
   t: TFunction
 }) => {
-  const typeLabel = (t?: string) => t ? addButtonGroupValueFormat[t].label : ''
-
   return (
     <Button
       size='sm'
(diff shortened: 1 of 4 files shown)


src/components/Header/AddButtonGroup.tsx
```diff
@@ -29,34 +29,31 @@ const addButtonTypes = ['core/planning-item', 'core/event', 'core/article', 'cor
 type Variant = VariantProps<typeof buttonVariants>['variant']
 type ButtonView = { name: View, type: string, icon?: { icon?: LucideIcon, color?: string } }
 
+const getViewLabel = (view: ButtonView, hast?: boolean) =>
+  view.name === 'Flash' && hast ? 'HAST' : (view.type ? addButtonGroupValueFormat[view.type].label : '')
+
```
gularsson: Maybe
```suggestion
  const getViewLabel = (view: ButtonView, hast?: boolean): string => {                                                                                                                                                                                                                                                                                                                                                                  
    if (view.name === 'Flash' && hast) {
      return 'HAST'                                                                                                                                                                                                                                                                                                                                                                                                                     
    }                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                                                                                                                                        
    return addButtonGroupValueFormat[view.type]?.label ?? ''                                                                                                                                                                                                                                                                                                                                                                            
  }  

```
To catch wrong keys, no nested ternaries and shorter line length

gularsson (approved):

dannelundqvist (approved):

src/components/Header/AddButtonGroup.tsx
vilindh: Fixed

gularsson (approved):

PR #2515, opened 2026-04-29
fix(assignments): disable action menu open when deliverable missing

gularsson:
Show a type-aware label (Öppna bild/video/flash etc.) and disable the item when the row has no deliverable uuid — opening with an empty id landed the Editor on a missing-document error.

### DIFF
--- a/src/views/Assignments/AssignmentColumns.tsx
+++ b/src/views/Assignments/AssignmentColumns.tsx
@@ -385,7 +385,7 @@ export function assignmentColumns<Ns extends Namespace>({ authors = [], locale, 
           ? t('common:actions.openType', { type: typeLabel })
           : t('views:assignments.actionMenu.openArticle')
         return (
-          <div className='shrink p-'>
+          <div className='shrink'>
             <DotMenu
               items={[
                 {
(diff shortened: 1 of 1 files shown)


vilindh (changes requested):
Would look nicer if the disabled item had an icon as well, plus the align would sync.
Perhaps `BanIcon` ?

src/views/Assignments/AssignmentColumns.tsx
```diff
@@ -373,22 +373,41 @@ export function assignmentColumns<Ns extends Namespace>({ authors = [], locale,
...
+          'document.meta.core_assignment.meta.core_assignment_type.value'
+        ]?.values?.[0]
+        const typeLabel = assignmentTypeValue
+          ? t(`shared:assignmentTypes.${assignmentTypeValue}` as TranslationKey)
+          : undefined
+        const openDeliverableLabel = typeLabel
+          ? t('common:actions.openType', { type: typeLabel })
+          : t('views:assignments.actionMenu.openArticle')
         return (
           <div className='shrink p-'>
```
dannelundqvist: Not part of this PR, but didn't see it before. Worth cleaning up the faulty class name `p-`.

dannelundqvist (approved):

vilindh (approved):

PR #2527, opened 2026-05-04
Quick-article: add preamble text block to template

vilindh:


dannelundqvist (approved):

src/components/QuickDocument/DialogEditor.tsx
```diff
@@ -35,7 +35,7 @@ export const DialogEditor = ({ ydoc, setTitle, onValidation, validateStateRef, t
     return <></>
   }
 
-  const countCharacters = type === 'flash' || type === 'hast'
+  const countCharacters = type === 'flash' || type === 'hast' || type === 'article'
```
gularsson: `type` is typed as `'article' | 'flash' | 'hast'`

according to that we could do:
```suggestion
const countCharacters = ['heading-1', 'preamble', 'body']
```
We're never ending up in else-ternary-branching?

🤔

gularsson (approved):
Did we want to featureFlag this or are we doing it for all?

src/components/QuickDocument/DialogEditor.tsx
vilindh: Correct, it's only ever going to end up there.
I'll fix it.

gularsson (approved):

PR #2528, opened 2026-05-04
fix(timeless): stop useDeliverableInfo refetch storm

gularsson:
The Timeless view fired ~150 GetDeliverableInfo calls per 20s of idle and produced 1.3s long tasks because every mounted instance of useDeliverableInfo unconditionally revalidated on every core/planning-item event, and TimelessRowActions mounted the hook once per row. Filter events by the linked planning UUID and defer the per-row fetch until the dot menu is opened. Same lazy-mount applied to Latest. After: 0 GetDeliverableInfo calls in 20s idle, worst long task 229ms.


dannelundqvist (approved):
👍

PR #2532, opened 2026-05-05
Feature/elelog 633 loading factboxes is slow

dannelundqvist:


gularsson (approved):
Perf improvement suggestion: Factbox query and Article query have no dependency on each other, we could run these in parallel and save some time (unclear how much

PR #2533, opened 2026-05-05
fix: remove onRowSelected callback

vilindh:
The onRowSelected callback effectively prevented Table items from being opened, if not provided from the parent component. As no current handler did anything other than to log the current row selected, this should be removed.


gularsson (approved):
Awesome 🧹

PR #2542, opened 2026-05-06
rework service URL resolution

hugowetterberg:
## Summary

- Public URLs derive from `BASE_PUBLIC_API_URL` (`https://<name>.<host>`); sparse `<NAME>_PUBLIC_URL` overrides supported.
- Internal URLs derive from `elephant-<name>` (overridable via `INTERNAL_API_NAMES`) on http:1080 or https:1443 (`INTERNAL_HTTPS_FOR_SERVICES`, with `*` opting in everything).
- `USE_PUBLIC_APIS_INTERNALLY=true` routes server-side calls through the public URLs — handy for local dev without cluster DNS.
- Resolver loads once at startup, validates input, and is injected via `RouteContext` server-side / `ServerConfig.resolveServiceUrl` on the client. No `process.env` reads outside startup.
- `/api/envs` now sends only `basePublicApiUrl` plus the sparse overrides map; the client memoizes URLs on demand so `useEffect` deps stay stable.
- Renames `spellcheckUrl` → `spellUrl` end to end.
- Folds NTB into the resolver: the translate route resolves it via `serviceUrls.resolveServiceUrl('ntb', 'internal')` instead of reading `NTB_URL` directly. Image search is unchanged — still constructs its NTB client from `imageSearchUrl`.
- README documents every consumed env var, including pattern variables (`<NAME>_PUBLIC_URL`, `INTERNAL_API_NAMES`, `INTERNAL_HTTPS_FOR_SERVICES`).

## Deployment notes

Breaking config change: any environment still setting `INDEX_URL`, `REPOSITORY_URL`, `USER_URL`, `SPELLCHECK_URL`, or `NTB_URL` directly needs to switch to `BASE_PUBLIC_API_URL` (+ overrides where required). Coordinate with the deploy configs before merge.


hugowetterberg:
Outdated

PR #2543, opened 2026-05-06
Call NTB nynorsk translation directly from the browser

dannelundqvist:
The BFF route at /api/translate could not reach the NTB cluster-internal address in deployed environments, surfacing as a 502 ("Translation service unreachable: fetch failed") in server logs. The route was a pure pass-through; moving the call to the browser removes the BFF hop entirely.

**Minimal changes. not taking new PR https://github.com/ttab/elephant-chrome/pull/2542 into account**

- Expose ntbUrl on /api/envs (NTB_PUBLIC_URL ?? NTB_URL ?? ''), optional
- Plumb through getServerEnvs → RegistryProvider → WireViewContent →
  createArticle → translateWireContent → translate()
- shared/translate.ts now POSTs directly to
  ${ntbUrl}/twirp/ttab.ntb.Nynorsk/Translate with the user's access token
- Delete src-srv/api/translate/index.ts
- Tests updated for the new options-bag signature


hugowetterberg (approved):

vilindh (approved):

__tests__/utils/envsEndpoint.test.ts
gularsson, line 11: We need to set `process.env.NTB_URL` here otherwise it'll read from dev-envs and fail locally.

gularsson (changes requested):

gularsson (approved):

PR #2567, opened 2026-05-11
Switch to workflow status handling of factboxes

dannelundqvist:


src/views/Factbox/index.tsx
```diff
@@ -270,8 +270,20 @@ const FactboxWrapper = (props: ViewProps & { documentId: string, data?: EleDocum
...
+  const handleRepositoryEvent = useCallback(
+    (event: { uuid?: string, mainDocument?: string }) => {
+      if (event.uuid === props.documentId || event.mainDocument === props.documentId) {
+        setStatusRefetchKey((k) => k + 1)
+      }
+    },
+    [props.documentId]
+  )
+
+  useRepositoryEvents(['core/factbox', 'core/factbox+meta'], handleRepositoryEvent)
```
gularsson: New reference on every render. Lift to a stable reference to avoid unsubscribe/resubscribe. `DocumentHistory` does it correctly with `eventTypes`.

src/views/Factbox/index.tsx
```diff
@@ -285,7 +297,7 @@ const FactboxWrapper = (props: ViewProps & { documentId: string, data?: EleDocum
         setDocumentState(getDocumentState(item))
       }
     })
```
gularsson: I think we need better error handling here, looking at `DocumentHistory` as an example

gularsson (changes requested):

gularsson (approved):

vilindh (approved):

PR #2586, opened 2026-05-13
fix(factbox): present usable status as published for nb locale

gularsson:
- nb: rename factbox usable title/description and draft/unpublished->usable transitions to publish wording
- workflowSpecification: drop factbox-specific bg-done usable color when i18n.language is nb
- DocumentHistory: keep factbox usable as usable (not done) when i18n.language is nb


vilindh (commented):
Instead of using the locale here, perhaps we could use the `imageSearchProvider` which is more specific for now

dannelundqvist (approved):

vilindh (approved):

PR #2587, opened 2026-05-13
Factbox: add LinkedArticles indicators

vilindh:


src/views/Factbox/lib/LinkedArticles.tsx
```diff
@@ -30,18 +48,74 @@ export const LinkedArticles = ({ documentId }: { documentId: string }): JSX.Elem
...
+    () => fetchDocuments<HitV1, typeof articleFields>({
       index: index!,
       session: session!,
       documentType: 'core/article',
-      fields,
-      query,
+      fields: articleFields,
+      query: articleQuery,
       sort: [{ field: 'modified', desc: true }],
       size: 100
```
gularsson: This seems like a _large_ amount of articles fetched...

src/views/Factbox/lib/LinkedArticles.tsx
```diff
@@ -30,18 +48,74 @@ export const LinkedArticles = ({ documentId }: { documentId: string }): JSX.Elem
...
+          .map((entry) => entry.values?.[SUBSET_NAME]?.block)
+          .filter((b): b is Block => b !== undefined)
+
+        result.set(item.uuid, blocks)
+      }
+
+      return result
+    }
+  )
+
```
gularsson: Do we really need the bulkget here? Can't we derive comparable information just from the query...

Or even more simple compare `rel.source.uuid`, if not current version then it's not the same...

I think this could be simplified... especially since we're in a boolean state of mind, changed or not changed... not what has changed...

gularsson (changes requested):

src/views/Factbox/lib/LinkedArticles.tsx
vilindh: Made a new commit, simplified

src/views/Factbox/lib/LinkedArticles.tsx
vilindh: Right. Perhaps worth adding a small pagination for this

dannelundqvist (approved):

PR #2598, opened 2026-05-18
fix(redis): exit on cache reconnect exhaustion

gularsson:
Add optional onUnrecoverable callback to the Redis class. The reconnectStrategy invokes it once, deferred via queueMicrotask, before returning its terminal Error after 10 failed attempts. Wire the callback in index.ts to log fatal and process.exit(1).

Refs ELELOG-657


hugowetterberg (approved):
Looks great!

PR #2605, opened 2026-05-19
Fix wires so that it if necessary fails visibly during article creation

dannelundqvist:


src/views/WireCreation/lib/createArticle.tsx
```diff
@@ -135,10 +135,15 @@ export async function createArticle({
...
   const localDate = convertToISOStringInTimeZone(dt, timeZone).slice(0, 10)
 
-  // Translate before any persistence. If the user opted into translation and
-  // it fails, abort the whole flow.
+  // Translate before any persistence; abort on failure so we never save
+  // an empty / untranslated article under a success toast.
   let articleContent: TBElement[] | undefined
-  if (wireContent && translationMode) {
+  if (translationMode) {
+    if (!wireContent) {
```
gularsson: Maybe even more hardening?
```suggestion
    if (!wireContent || wireContent.length === 0) {
```
Check for content as well?

gularsson (approved):

PR #2606, opened 2026-05-19
Feature/elelog 667 wirecreation articleid may regenerate mid dialog

dannelundqvist:


src/views/WireCreation/index.tsx
gularsson: checking `typeof formId === 'string' && typeof articleId === 'string'` is not needed now

gularsson (approved):

src/views/WireCreation/index.tsx
dannelundqvist: True!

gularsson (approved):

vilindh (approved):

PR #2607, opened 2026-05-19
fix: embedded factbox at out of range index hangs the loader forever

vilindh:


src/views/Factbox/index.tsx
```diff
@@ -188,6 +196,15 @@ const EmbeddedFactboxView = (props: ViewProps & { data?: EleDocumentResponse }):
     }
   }
 
+  if (props.hasError) {
+    return (
+      <ErrorView
+        title={t('errors:toasts.couldNotOpenEmbeddedFactbox')}
+        message={t('errors:messages.embeddedFactboxMissing')}
```
gularsson, line 203: 3 failure paths, one error message - `The article no longer contains a factbox at this position`

This message works for `!block` check, not the other two, either make it more generic `Couldn't open factbox` or more suiting messages?

gularsson (approved):

gularsson (approved):

PR #2610, opened 2026-05-19
fix: factbox status fetch error leaves UI in undefined state

vilindh:


src/views/Factbox/index.tsx
```diff
@@ -271,6 +271,8 @@ const FactboxWrapper = (props: ViewProps & { documentId: string, data?: EleDocum
   const [documentState, setDocumentState] = useState<DocumentState | undefined>(undefined)
   const [currentVersion, setCurrentVersion] = useState<bigint | undefined>(undefined)
   const [statusRefetchKey, setStatusRefetchKey] = useState(0)
+  const [statusFetchState, setStatusFetchState] = useState<'idle' | 'loading' | 'error'>('idle')
```
dannelundqvist: `statusFetchState` is set to `'loading'` on line lin 297, but the state is never used to differentiate the rendering which is a bit confusing. Is it really needed? Or is there something we would want to render differently while loading compared to idle or error?

dannelundqvist (approved):
It works but is a bit confusing with the 'loading' state never used and the useEffect that is a "passivively" triggered by increasing the statusRefetchKey. Not sure if it could have been turned into a proper callback (or if it is worth it now). So setting a white check anyways.

src/views/Factbox/index.tsx
vilindh: No, a stray piece of code from an earlier attempt.
I'll refactor some of the code, based on your comment above.

gularsson (approved):

PR #2618, opened 2026-05-19
Localize missing section message

dannelundqvist:


src/locales/nb/wires.json
```diff
@@ -58,6 +58,7 @@
     "articleCreated": "Artikkel opprettet",
     "createError": "Kunne ikke opprette en artikkel!",
     "createError2": "Kunne ikke opprette ny artikkel!",
+    "planningMissingSection": "Valgt plan mangler seksjon og kan ikke brukes.",
```
vilindh: Looking at https://github.com/ttab/elephant-chrome/blob/main/src/locales/nb/core.json, nb localization of section is "kategori"

vilindh (changes requested):

src/locales/nb/wires.json
dannelundqvist: It was inconsistent. I fixed two more places as part of this.

vilindh (approved):

PR #2636, opened 2026-05-22
It should be possible to clear the translation settings. When cleared…

dannelundqvist:
… only standard should be an option, otherwise personal prefs.


vilindh (approved):
✅

gularsson (approved):

PR #2639, opened 2026-05-22
Feature/elelog 698 print article image upload shows success toast but no image

gularsson:


src/views/PrintEditor/lib/getImageSrc.ts
```diff
@@ -0,0 +1,40 @@
...
+    return (properties.src as string) || ''
+  }
+
+  const session = await getCachedSession()
+  if (!session?.accessToken) {
+    console.warn('getImageSrc: no cached session, cannot resolve attachment', { uploadId })
+    return ''
+  }
+
+  let details
```
dannelundqvist: I would probably ditch the `let` declaration and go for something like below (not tested):
```
try {
  const details = await repository.getAttachmentDetails(uploadId, session.accessToken)

  if (!details?.downloadLink) {
    console.warn('getImageSrc: attachment has no downloadLink', { uploadId })
    return ''
  }
  return details.downloadLink
} catch (ex) {
  console.error('getImageSrc: getAttachmentDetails failed', { uploadId, ex })
  throw ex
}
```

Then again. It is not that important.

dannelundqvist (approved):

dannelundqvist (approved):

PR #2641, opened 2026-05-22
fix(editor): show platform-aware keyboard shortcut legend

gularsson:
Textbit's built-in hotkey renderer hardcodes Mac glyphs (⌘ ⌥ ⌃ ⇧) for every platform, so Windows, Linux and Chromebook editors see opaque labels like ⌘⌥1 next to menu items even though the shortcut they need to press is Ctrl+Alt+1.

Introduce a small formatter and pass platform-aware children into Menu.Hotkey. Mac users keep the Unicode glyphs; everyone else now sees the spelled-out modifier names joined by "+", e.g. "Ctrl+Alt+1", "Ctrl+B", "Ctrl+Alt+Up".

Refs ELELOG-624

### DIFF
--- a/__tests__/formatHotkey.test.ts
+++ b/__tests__/formatHotkey.test.ts
@@ -1,0 +1,100 @@
+import { describe, it, expect, vi, afterEach } from 'vitest'
+import { formatHotkey } from '@/lib/formatHotkey'
+
+describe('formatHotkey', () => {
+  describe('on Mac', () => {
+    const isMac = true
+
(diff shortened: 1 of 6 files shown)


dannelundqvist:
Should this maybe be fixed in Textbit where this logic is today? https://github.com/ttab/textbit/blob/4a929e2c92eddb9e5d8ddbefc3a6d8b0833c9315/lib/utils/modifier.ts?

PR #2655, opened 2026-05-29
feat(imagesearch): add NTB distributor filter

gularsson:
- bump @ttab/elephant-tt-api 0.4.2 -> 0.5.2
- read distributorNames from the SWR key in ntbFetcher and set outsideSubscription when a distributor is selected
- add distributor Select to ImageSearchInput, shown only for NTB
- default NPK-unit users to the NPK distributor via useHasUnit
- introduce shared ImageSearchKey tuple type used by both fetchers
- add distributor/allDistributors labels (sv, en, nb)
- cover the fetcher coupling and unit-driven default with tests

### DIFF
--- a/src/views/ImageSearch/index.tsx
+++ b/src/views/ImageSearch/index.tsx
@@ -9,10 +9,11 @@ import { Thumbnail } from './Thumbnail'
 import { ImageSearchInput } from './SearchInput'
 import { SWRConfig } from 'swr'
 import { createTTFetcher } from './lib/ttFetcher'
-import { createNTBFetcher } from './lib/ntbFetcher'
+import { createNTBFetcher, NPK_DISTRIBUTOR, NPK_UNIT, type NTBDistributor } from './lib/ntbFetcher'
 import { useRegistry } from '@/hooks/useRegistry'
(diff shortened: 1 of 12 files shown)


vilindh (approved):

src/views/ImageSearch/index.tsx
```diff
@@ -86,13 +87,17 @@ const ImageSearchContent = ({
   mediaType: MediaTypes
   isNtb: boolean
 }): JSX.Element => {
+  const isNpkUser = useHasUnit(NPK_UNIT)
   const [queryString, setQueryString] = useState('')
+  const [distributorNames, setDistributorNames] = useState<NTBDistributor[]>(
```
dannelundqvist, line 92: Not sure but, if `ImageSearchContent` mounts before the the session resolve (`isNpkUser` is `false`) the state will stay `[]` forever?

Maybe we could guard the mount on `status === 'authenticated'` before rendering `ImageSearchContent`.

dannelundqvist (changes requested):

dannelundqvist (approved):

PR #2659, opened 2026-06-01
Handle nynors translation settings values and multiple values

dannelundqvist:


hugowetterberg (approved):

gularsson (commented):
Works

Future improvement: use Twirp client instead of direct fetch. Removes the use for rolling our own types (twice) and better unity with for example imagesearch

